PR #12, opened 2015-02-06
Dependecies update

ilia-sheremet:


pavel-kirienko:
There's an extra file in this PR, please remove it.

PR #20, opened 2015-04-02
Prepend pyuavcan in source directory to sys.path

bendyer:
When running from the source directory, libuavcan_dsdlc should
use the version of pyuavcan in the source directory instead of
a globally installed version of pyuavcan.

Fixes UAVCAN/uavcan#18


pavel-kirienko:
Thanks Ben

PR #24, opened 2015-04-24
Has issue will compiling

davids5:
uavcan/libuavcan/src/driver/uc_can.cpp

std.hpp:70:37: error: 'std::size_t' has not been declared  
extern int snprintf(char out, **std::size_t** maxlen, const char

Not sure if this is the "correct" fix, but it works


pavel-kirienko:
Thanks David.

PR #27, opened 2015-05-18
Posix tools refactor

pavel-kirienko:
@davids5 Please review the changes, and merge if everything looks OK.

- The file structure I proposed initially was inconsistent with other drivers, which I now fixed.
- Linux driver now depends on the POSIX driver and includes a simple test application for it.
- Format string of event tracer has been fixed; also commas were replaced with tabs for better readability.


davids5:
All looks good - I will keep changes in mind for next submission - thank you!

PR #39, opened 2015-06-10
libuavcan: Disable Frame Listener with UAVCAN_TINY

kylemanna:
* Compliation will fail on small systems with UAVCAN_TINY defined
  with the following error:

      abstract_node.hpp:123:33: error: 'IRxFrameListener' has not been declared

* Resolve issue by removing unecessary functions.
* Error is revealed and resolved when building test_stm32f107.


pavel-kirienko:
Thanks Kyle!

PR #41, opened 2015-06-15
Can HW filter feature + getIface() corrections 

ilia-sheremet:


libuavcan/include/uavcan/transport/can_acceptance_filter_configurator.hpp
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (C) 2014 Pavel Kirienko <pavel.kirienko@gmail.com>,
+ *                    Ilia  Sheremet <illia.sheremet@gmail.com>
+ */
+
+#ifndef CAN_ACCEPTANCE_FILTER_CONFIGURATOR_HPP_INCLUDED
+#define CAN_ACCEPTANCE_FILTER_CONFIGURATOR_HPP_INCLUDED
```
pavel-kirienko: Proper guard name would be `UAVCAN_TRANSPORT_CAN_ACCEPTANCE_FILTER_CONFIGURATOR_HPP_INCLUDED` (the current one lacks prefix).

libuavcan/include/uavcan/build_config.hpp
```diff
@@ -228,6 +228,17 @@ static const unsigned FloatComparisonEpsilonMult = UAVCAN_FLOAT_COMPARISON_EPSIL
 static const unsigned FloatComparisonEpsilonMult = 10;
 #endif
 
+/**
+ * Maximum number of CAN acceptance filters available on a platform
```
pavel-kirienko: > on a platform

I guess you meant on **the** platform? :)

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
+/*
```
pavel-kirienko: Please rename the file to `uc_can_acceptance_filter_configurator.cpp` (all other `.cpp` files have prefixed names, this is needed to avoid name collisions on certain build systems).

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
```diff
@@ -601,6 +601,10 @@ class SocketCanDriver : public uavcan::ICanDriver
     {
         return (iface_index >= num_ifaces_) ? nullptr : static_cast<SocketCanIface*>(ifaces_[iface_index]);
     }
+	const SocketCanIface* getIface(std::uint8_t iface_index) const override
+	{
+		return (iface_index >= num_ifaces_) ? nullptr : static_cast<const SocketCanIface*>(ifaces_[iface_index]);
+	}
```
pavel-kirienko: Please expand tabs.

libuavcan/test/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,156 @@
...
+	uint32_t configure_array_size = configure_array.getSize();
+
+    ASSERT_EQ(configure_filters_assert, 0);
+    ASSERT_EQ(configure_array_size, 4);
+
+		for (uint16_t i=0; i<configure_array_size; i++)
+		{
+			std::cout<<"config.ID ["<<i<<"]= "<< configure_array.getByIndex(i)-> id<<std::endl;
+			std::cout<<"config.MK ["<<i<<"]= "<< configure_array.getByIndex(i)-> mask<<std::endl;
+		}
```
pavel-kirienko: Please expand tabs.

libuavcan/include/uavcan/driver/can.hpp
```diff
@@ -85,6 +85,16 @@ struct UAVCAN_EXPORT CanFilterConfig
 {
     uint32_t id;
     uint32_t mask;
+
+    bool operator == (const CanFilterConfig& rhs) const
```
pavel-kirienko: Did uncrustify put spaces around `==`? If so, sorry, I'll try to teach it otherwise. If not, please remove the spaces ().

libuavcan/test/node/test_node.hpp
```diff
@@ -74,6 +75,15 @@ struct PairableCanDriver : public uavcan::ICanDriver, public uavcan::ICanIface
...
     }
 
+    virtual const uavcan::ICanIface* getIface(uavcan::uint8_t iface_index) const
+    {
+        if (iface_index == 0)
+        {
+            return this;
+        }
+        return NULL;
+    }
```
pavel-kirienko: This overload is no longer needed, because there's a default virtual implementation in the base class that does the same thing ([line 180](https://github.com/UAVCAN/libuavcan/pull/41/files#diff-dfb72932e25cf3185462fb93974efa06R180)).

libuavcan/test/transport/can/can.hpp
```diff
@@ -232,6 +232,7 @@ class CanDriverMock : public uavcan::ICanDriver
     }
 
     virtual uavcan::ICanIface* getIface(uavcan::uint8_t iface_index) { return &ifaces.at(iface_index); }
+    virtual const uavcan::ICanIface* getIface(uavcan::uint8_t iface_index) const { return &ifaces.at(iface_index); }
```
pavel-kirienko: Same as above, this implementation is now redundant.

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
pavel-kirienko: Just noted that this implementation is redundant, as I also mentioned in two places above.

libuavcan_drivers/lpc11c24/driver/include/uavcan_lpc11c24/can.hpp
```diff
@@ -61,6 +61,8 @@ class CanDriver
 
     virtual uavcan::ICanIface* getIface(uavcan::uint8_t iface_index);
 
+    virtual const uavcan::ICanIface* getIface(uavcan::uint8_t iface_index) const;
```
pavel-kirienko: Ditto.

libuavcan/include/uavcan/transport/can_acceptance_filter_configurator.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include <uavcan/error.hpp>
+#include <uavcan/transport/dispatcher.hpp>
+#include <uavcan/node/abstract_node.hpp>
+#include <uavcan/build_config.hpp>
+#include <uavcan/util/multiset.hpp>
+
+namespace uavcan
+{
+
+class CanAcceptanceFilterConfigurator
```
pavel-kirienko: A brief comment describing what it is and how to use it would help here.

See an example here: https://github.com/UAVCAN/libuavcan/blob/master/libuavcan/include/uavcan/protocol/node_info_retriever.hpp#L61-L90

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+            UAVCAN_ASSERT(0);
+            out = 0;
+            break;
+        }
+        const uint16_t num = iface->getNumFilters();
+        out = min(out, num);
+        if (num > MaxCanAcceptanceFilters)
+        {
+            return MaxCanAcceptanceFilters;
+        }
```
pavel-kirienko: Ha, a bug!

The objective is to find the lowest number of HW filters. The function is supposed to traverse all interfaces and find the lowest value.

The condition here breaks the logic if one of the interfaces has more filters than `MaxCanAcceptanceFilters`. For example, if there are three interfaces with 5, 6, and 7 filters, and `MaxCanAcceptanceFilters` is 7, the function will return 7 instead of 5.

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+        return -ErrLogic;
+    }
+
+    for (uint16_t i = 0; i < filter_array_size; i++)
+    {
+        CanFilterConfig temp_filter_config = *multiset_configs_.getByIndex(i);
+        if (multiset_configs_.getByIndex(i) == NULL)
+        {
+            UAVCAN_ASSERT(0);
+        }
```
pavel-kirienko: UAVCAN_ASSERT(multiset_configs_.getByIndex(i) != NULL);

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+        if ((temp_filter_config.id != 0) || (temp_filter_config.mask != 0))
```
pavel-kirienko: This condition can accept a config where `mask` is zero, which is not a valid config. I suggest to either change the condition to check only mask, or to add `UAVCAN_ASSERT(temp_filter_config.mask != 0);`. The first solution makes more sense, but maybe I just failed to understand your intent.

libuavcan/include/uavcan/transport/can_acceptance_filter_configurator.hpp
```diff
@@ -0,0 +1,90 @@
...
+    typedef uavcan::Multiset<CanFilterConfig, 1> MultisetConfigArray;
+
+    static CanFilterConfig mergeFilters(CanFilterConfig &a_, CanFilterConfig &b_);
+    static uint8_t countBits(uint32_t n_);
+    uint16_t getNumFilters() const;
+
+    /**
+     * Fills the multiset_configs_ to proceed it with computeConfiguration()
+     */
+    int16_t fillArray();
```
pavel-kirienko: If it fills a multiset, why is it named `fillArray()`? :)

How about `loadInputConfiguration()` or whatever.

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        *multiset_configs_.getByIndex(j_rank) = mergeFilters(*multiset_configs_.getByIndex(i_rank),
+                                                             *multiset_configs_.getByIndex(j_rank));
+        CanFilterConfig temp_key = *multiset_configs_.getByIndex(i_rank);
+
+        if ((multiset_configs_.getByIndex(i_rank)->id == multiset_configs_.getByIndex(j_rank)->id) &&
+            (multiset_configs_.getByIndex(i_rank)->mask == multiset_configs_.getByIndex(j_rank)->mask))
+        {
+            multiset_configs_.removeAll(temp_key);             // If i_rank and j_rank elements are the same, we need to add it
```
pavel-kirienko: Line length limit exceeded here (120 chars).

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+const unsigned CanAcceptanceFilterConfigurator::DefaultFilterServiceRequestMask;
+const unsigned CanAcceptanceFilterConfigurator::ServiceRespFrameID;
+const unsigned CanAcceptanceFilterConfigurator::ServiceRespFrameMask;
+
+int16_t CanAcceptanceFilterConfigurator::fillArray()
+{
+    multiset_configs_.clear();
+    uint16_t i = 0;
+
+    CanFilterConfig ServiseRespFrame;
```
pavel-kirienko: Improper coding style (`service_resp_frame`).

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+    ServiseRespFrame.id = ServiceRespFrameID;
+    ServiseRespFrame.mask = ServiceRespFrameMask;
+    multiset_configs_.emplace(ServiseRespFrame);
+    if (multiset_configs_.getByIndex(i) == NULL)
+    {
+        UAVCAN_ASSERT(0);
+    }
```
pavel-kirienko: ```UAVCAN_ASSERT(multiset_configs_.getByIndex(i) != NULL);```

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+        multiset_configs_.emplace(cfg);
+        if (multiset_configs_.getByIndex(i) == NULL)
+        {
+            UAVCAN_ASSERT(0);
+        }
+
+        i++;
+        p1 = p1->getNextListNode();
+    }
```
pavel-kirienko: This loop basically replicates the one above, I suggest to extract it into a method.

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+        return -ErrDriver;
+    }
+
+    for (uint16_t i = 0; i < multiset_configs_.getSize(); i++)
+    {
+        UAVCAN_TRACE("CanAcceptanceFilterConfigurator::fillArray()", "map.ID [%u] = %d", i,
+                     multiset_configs_.getByIndex(i)->id);
+        UAVCAN_TRACE("CanAcceptanceFilterConfigurator::fillArray()", "map.MK [%u] = %d", i,
+                     multiset_configs_.getByIndex(i)->mask);
+    }
```
pavel-kirienko: I suggest to enclose this loop with `#if UAVCAN_DEBUG ... #endif`.

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+    return 0;
+}
+
+int16_t CanAcceptanceFilterConfigurator::computeConfiguration()
+{
+    uint16_t i_rank = 0, j_rank = 0;
```
pavel-kirienko: Scope of these variables can be reduced - move them into the loop below.

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+        UAVCAN_ASSERT(multiset_configs_.getSize() != 0);
+
+        uint8_t best_rank = 0;
+        const uint16_t map_array_size = static_cast<uint16_t>(multiset_configs_.getSize());
+
+        for (uint16_t i_ind = 0; i_ind < map_array_size - 1; i_ind++)
+        {
+            for (uint16_t j_ind = static_cast<uint16_t>(i_ind + 1); j_ind < map_array_size; j_ind++)
+            {
+                CanFilterConfig temp_array = mergeFilters(*multiset_configs_.getByIndex(i_ind),
```
pavel-kirienko: Why `temp_array`?

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,256 @@
...
+        if ((multiset_configs_.getByIndex(i_rank)->id == multiset_configs_.getByIndex(j_rank)->id) &&
+            (multiset_configs_.getByIndex(i_rank)->mask == multiset_configs_.getByIndex(j_rank)->mask))
+        {
+            multiset_configs_.removeAll(temp_key);             // If i_rank and j_rank elements are the same, we need to add it
+            multiset_configs_.emplace(temp_key);               //another time, to preclude removal of both elements
+        }
+        else
+        {
+            multiset_configs_.removeAll(temp_key);
+        }
```
pavel-kirienko: Isn't is possible to replace everything on lines 114...125 with just this:

```
multiset_configs_.removeFirst(*multiset_configs_.getByIndex(i_rank));
```

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
ilia-sheremet: I think this condition remains from the time when the algorithm was filling configurations with zeros, instead of deleting them. So I would just add UAVCAN_ASSERT(temp_filter_config.mask != 0);

pavel-kirienko: > theoretically mask might be equal to zero in case there are much more listeners than available filters, so it would merge until all digits go zero.

Valid point, though a configuration that has at least one zero mask is meaningless (it won't filter anything). It's not actually a problem of the algorithm itself, it merely means that there's no solution for the provided inputs. It's just an observation, I'm not proposing any changes.

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
ilia-sheremet: so it means that we could calculate configurations maximum for 256 listeners? Does it suit us?

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
ilia-sheremet: We can do that, it will make program more efficient in case there are no the same 3+ listeners (which is likely). But does
multiset_configs_.removeFirst(*multiset_configs_.getByIndex(i_rank));
delete that one particular element starting from the beginning of container? or any element with the same value?

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
pavel-kirienko: Uhm, why 256?

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
pavel-kirienko: It deletes only the first occurrence of an element with the same value (and also it may or may not shuffle some or all of other elements). Unless I'm terribly wrong, the algorithm doesn't care about which particular copy gets removed, so that should be safe.

libuavcan/src/transport/can_acceptance_filter_configurator.cpp
ilia-sheremet: nevermind, I thought you mean i and j, not i/j_rank

pavel-kirienko:
Ilia, when you're done with fixing, please merge master into your branch so we can merge it back automatically.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,230 @@
...
+const unsigned CanAcceptanceFilterConfigurator::ServiceRespFrameMask;
+
+int16_t CanAcceptanceFilterConfigurator::loadInputConfiguration()
+{
+    multiset_configs_.clear();
+
+    CanFilterConfig service_resp_cfg;
+    service_resp_cfg.id = ServiceRespFrameID;
+    service_resp_cfg.mask = ServiceRespFrameMask;
+	if (multiset_configs_.emplace(service_resp_cfg) == NULL)
```
pavel-kirienko: Bad formatting (did you run uncrustify on this file?)

Also, constructor initializer lists are still formatted weirdly (the humanity would surely benefit if somebody developed an adequate autoformatter capable of handling initializer lists without breaking things); above I found this:

```c++
CanSelectMasks() :
    read(0), write(0)
{ }
```

Which is supposed to be:

```c++
CanSelectMasks() :
    read(0),
    write(0)
{ }
```

It's quite a minor and insignificant issue, I just want to get to the bottom of it: who formatted it this way - you or uncrustify? :) If it's uncrustify, what version did this? I thought I wrote a more or less adequate config that should instruct it to handle constructors properly.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,230 @@
...
+                     multiset_configs_.getByIndex(i)->mask);
+    }
+#endif
+
+    return 0;
+}
+
+int16_t CanAcceptanceFilterConfigurator::computeConfiguration()
+{
+    uint8_t i_rank = 0, j_rank = 0;
```
pavel-kirienko: Why `uint8_t`? Looking at the code I suspect that it's supposed to be `uint16_t`.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,230 @@
...
+    if (filter_array_size > MaxCanAcceptanceFilters)
+    {
+        UAVCAN_ASSERT(0);
+        return -ErrLogic;
+    }
+
+    for (uint16_t i = 0; i < filter_array_size; i++)
+    {
+        CanFilterConfig temp_filter_config = *multiset_configs_.getByIndex(i);
+        UAVCAN_ASSERT(multiset_configs_.getByIndex(i) != NULL);
```
pavel-kirienko: If I was a rigorous static analysis tool, I'd say "Null check after dereferencing is meaningless".

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    ICanDriver& can_driver = node_.getDispatcher().getCanIOManager().getCanDriver();
+    for (uint8_t i = 0; i < node_.getDispatcher().getCanIOManager().getNumIfaces(); i++)
+    {
+        ICanIface* iface = can_driver.getIface(i);
+        if (iface == NULL)
+        {
+            UAVCAN_ASSERT(0);
+            break;
```
pavel-kirienko: Are you sure it's not supposed to be `return -ErrDriver`? Because the current code will return zero on failure.

libuavcan/test/transport/can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,157 @@
...
+        DataType msg;
+
+        ReceivedDataStructureCopy(const ReceivedDataStructure& s)
+        : ts_monotonic(s.getMonotonicTimestamp())
+        , ts_utc(s.getUtcTimestamp())
+        , transfer_type(s.getTransferType())
+        , transfer_id(s.getTransferID())
+        , src_node_id(s.getSrcNodeID())
+        , iface_index(s.getIfaceIndex())
+        , msg(s)
```
pavel-kirienko: Uncrustify?

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+int16_t CanAcceptanceFilterConfigurator::computeConfiguration()
+{
+    uint8_t i_rank = 0, j_rank = 0;
+    const uint16_t acceptance_filters_number = getNumFilters();
+
+    while (acceptance_filters_number < multiset_configs_.getSize())
+    {
+        UAVCAN_ASSERT(multiset_configs_.getSize() != 0);
```
pavel-kirienko: (static analyzer hat on): This check contradicts the loop termination condition. Probably you wanted to put it before the loop?

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
ilia-sheremet: It's uncrustify who does that. Apart from this it makes spaces like this:
uint8_t dlc;                // /< Data Length Code
if there are more than two slashes it makes space after them.
I think I have the latest config from Zubax. Here is the one I use:
https://drive.google.com/file/d/0B9wE0Um8rdPgTUFwUGZmQjd2T2c/view?usp=sharing

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
ilia-sheremet: Last time you wrote to "Scope of these variables can be reduced", I misunderstood as to reduce the memory of type. Now I got it, corrected and put in While loop.

libuavcan/test/transport/can_acceptance_filter_configurator.cpp
ilia-sheremet: almost =) that's how uncrustify do:
        ReceivedDataStructureCopy(const ReceivedDataStructure& s) :
            ts_monotonic(s.getMonotonicTimestamp()), ts_utc(s.getUtcTimestamp()), transfer_type(s.getTransferType()),
            transfer_id(s.getTransferID()), src_node_id(s.getSrcNodeID()), iface_index(s.getIfaceIndex()), msg(s)
        { }
just putting everything in 2 lines. My bad, I corrected it a bit wrong

PR #45, opened 2015-06-26
Can HW acceptance filter feature + getIface() corrections + dsdl hash fixed

ilia-sheremet:


libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,228 @@
...
+    while (p)
+    {
+        CanFilterConfig cfg;
+        cfg.id = static_cast<uint32_t>(p->getDataTypeDescriptor().getID().get()) << 16;
+        cfg.id |= static_cast<uint32_t>(p->getDataTypeDescriptor().getKind()) << 8;
+        cfg.mask = DefaultFilterMsgMask;
+        if (multiset_configs_.emplace(cfg) == NULL)
+            {
+                return -ErrMemory;
+            }
```
pavel-kirienko: Formatting.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,228 @@
...
+const unsigned CanAcceptanceFilterConfigurator::ServiceRespFrameMask;
+
+int16_t CanAcceptanceFilterConfigurator::loadInputConfiguration()
+{
+    multiset_configs_.clear();
+
+    CanFilterConfig service_resp_cfg;
+    service_resp_cfg.id = ServiceRespFrameID;
+    service_resp_cfg.mask = ServiceRespFrameMask;
+	if (multiset_configs_.emplace(service_resp_cfg) == NULL)
```
pavel-kirienko: Formatting.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,228 @@
...
+                     multiset_configs_.getByIndex(i)->mask);
+    }
+#endif
+
+    return 0;
+}
+
+int16_t CanAcceptanceFilterConfigurator::computeConfiguration()
+{
+    const uint16_t acceptance_filters_number = getNumFilters();
```
pavel-kirienko: If `getNumFilters()` returns zero, the algorithm will fail. This corner case should be checked before the loop starts.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -0,0 +1,228 @@
...
+            return -ErrDriver;
+        }
+    }
+
+    return 0;
+}
+
+int CanAcceptanceFilterConfigurator::configureFilters()
+{
+    if (getNumFilters() == 0)
```
pavel-kirienko: This check is nice, but it doesn't protect against transient failures, so `computeConfiguration()` should check it too.

pavel-kirienko: Alternatively, you could store the number of filters there, and pass the value into `computeConfiguration()`.

pavel-kirienko:
Спасибо.

ilia-sheremet:
Всегда пожалуйста =)

PR #52, opened 2015-07-07
Recovered build

davids5:
DONOT MERGE

@pavel-kirienko FYI: @LorenzMeier 

The changes made on the  new_transport_layer branch broke the PX4 Build.

There was a major change in the interfaces and the stm32 driver used in PX4 was not even built.

int BusEvent::poll(::file filp, ::pollfd fds, bool setup) - will not compile and it is not clear how to gain the data needed in new definition of makeSelectMasks()

uavcan::CanSelectMasks CanDriver::makeSelectMasks(const uavcan::CanFrame* (& pending_tx)[uavcan::MaxCanIfaces])


pavel-kirienko:
The driver is still a work in progress; as you already observed, I didn't even build it for px4, so any update now would be very premature.

PR #61, opened 2015-08-21
Add UAVCAN_NO_GLOBAL_DATA_TYPE_REGISTRY flag

bendyer:
This allows the global data type registry to be disabled if it's not needed, saving 2–4 KiB depending on compilation settings and the data types used.

### DIFF
--- a/libuavcan/include/uavcan/build_config.hpp
+++ b/libuavcan/include/uavcan/build_config.hpp
@@ -104,7 +104,7 @@
  * Disable the global data type registry, which can save some space on embedded systems.
  */
 #ifndef UAVCAN_NO_GLOBAL_DATA_TYPE_REGISTRY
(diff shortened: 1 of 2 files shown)


libuavcan/include/uavcan/build_config.hpp
```diff
@@ -101,6 +101,13 @@
 #endif
 
 /**
+ * Disable the global data type registry, which can save some space on embedded systems.
+ */
+#ifndef UAVCAN_NO_GLOBAL_DATA_TYPE_REGISTRY
+# define UAVCAN_NO_GLOBAL_DATA_TYPE_REGISTRY 1
```
pavel-kirienko: Default should be zero.

bendyer:
Indeed it should. Fixed.

PR #62, opened 2015-08-21
Master baremetal

davids5:


pavel-kirienko:
Tested on [TE-STM32F107](https://www.terraelectronica.ru/catalog_info.php?CODE=1023380) with ChibiOS, works flawlessly. Feel free to merge if you consider it finished.

davids5:
It is not the prettiest yet, but functional. We should do an IO consolidation pass - but that is not a priority

PR #67, opened 2015-10-03
Cmake cross

jgoppert:


CMakeLists.txt
```diff
@@ -4,10 +4,63 @@
...
+		if("${NUTTX_EXPORT_DIR}" STREQUAL "")
+			message(FATAL_ERROR "please define the nuttx export directory")
+		endif()
+	endif()
+	include_directories(${NUTTX_EXPORT_DIR}/include)
+	link_directories(${NUTTX_EXPORT_DIR}/lib)
+	link_libraries(
+		nosys
+		nuttx	
+		apps
```
pavel-kirienko: Isn't this set of libraries specific to PX4 configuration, and may not be valid for a generic NuttX build?

jgoppert: We could pass in a list of nuttx link libraries to be the most flexible.

jgoppert:
This is ready for review.

libuavcan/CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ include_directories(${DSDLC_OUTPUT})
 #
 if (COMPILER_IS_GCC_COMPATIBLE)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wundef")
-    if (USE_CPP03)
+if (UAVCAN_USE_CPP03)
```
jgoppert: need to fix indent

CMakeLists.txt
```diff
@@ -4,29 +4,130 @@
...
+#=============================================================================
+# set flags
+#
+
+add_definitions(
+    -DUAVCAN_MEM_POOL_BLOCK_SIZE=${UAVCAN_MEM_POOL_BLOCK_SIZE}
+    -DUAVCAN_MAX_NETWORK_SIZE_HINT=${UAVCAN_MAX_NETWORK_SIZE_HINT}
+    )
 
+set(bool_opts UAVCAN_NO_ASSERTIONS UAVCAN_USE_EXTERNAL_SNPRINTF)
```
jgoppert: need to make sure all bool arguments are added, looks like i'm missing: UAVCAN_USE_CPP03, but I think that is handled elsewhere.

.gitignore
```diff
@@ -22,3 +22,6 @@ dsdlc_generated
 
 # Log files
 *.log
+
+# cmake build directories
+build*/
```
pavel-kirienko: This is redundant, `build` is already listed above.

CMakeLists.txt
```diff
@@ -4,29 +4,131 @@
...
+    "UAVCAN_USE_CPP03:BOOL:OFF::Use cpp03 standard."
+    "UAVCAN_NO_ASSERTIONS:BOOL:OFF::Enable assertions."
+    "UAVCAN_USE_EXTERNAL_SNPRINTF:BOOL:OFF::Use external snprintf."
+    "UAVCAN_MEM_POOL_BLOCK_SIZE:STRING:48::memory pool block size"
+    "UAVCAN_STM32_TIMER_NUMBER:STRING:5::default timer"
+    "UAVCAN_MAX_NETWORK_SIZE_HINT:STRING:16::max network size"
+    "UAVCAN_PLATFORM:STRING:generic:generic linux stm32:Platform."
+    "UAVCAN_OS:STRING:generic:generic posix nuttx chibios:Operating system."
+    "UAVCAN_STM32_NUM_IFACES:STRING:2:1 2:Number of interfaces."
+    )
```
pavel-kirienko: Is it really necessary to duplicate options here? This violates DRY and requires extra maintenance. I suggest to keep using them via preprocessor, like we do with make.

CMakeLists.txt
```diff
@@ -4,29 +4,131 @@
...
+
+string(TOUPPER ${UAVCAN_OS} OS_UPPER)
+
+#=============================================================================
+# find libraries
+#
+if(${UAVCAN_OS} STREQUAL "nuttx")
+    find_path(NUTTX_EXPORT_DIR libs/libnuttx.a
+        PATHS
+        $ENV{HOME}/git/px4/Firmware/build_px4fmu-v2_default/px4fmu-v2/NuttX/nuttx-export
```
pavel-kirienko: Environment-specific path, this is wrong.

CMakeLists.txt
```diff
@@ -4,29 +4,131 @@
...
 
-project(uavcan)
+#=============================================================================
+# set flags
+#
+
+add_definitions(
+    -DUAVCAN_MEM_POOL_BLOCK_SIZE=${UAVCAN_MEM_POOL_BLOCK_SIZE}
+    -DUAVCAN_MAX_NETWORK_SIZE_HINT=${UAVCAN_MAX_NETWORK_SIZE_HINT}
+    )
```
pavel-kirienko: Same question as above - why can't we pass these via preprocessor definitions? The application itself can configure these via `add_definitions()`, globally. There will be no name clashing because all preprocessor definitions used by libuavcan are prefixed with `UAVCAN_`.

CMakeLists.txt
```diff
@@ -4,29 +4,131 @@
...
+            ${NUTTX_EXPORT_DIR}/include
+            ${NUTTX_EXPORT_DIR}/include/cxx
+            ${NUTTX_EXPORT_DIR}/arch/chip
+            ${NUTTX_EXPORT_DIR}/arch/common
+            )
+    endif()
+    add_definitions(
+        -DUAVCAN_STM32_${OS_UPPER}=1
+        -DUAVCAN_STM32_NUM_IFACES=${UAVCAN_STM32_NUM_IFACES}
+        -DUAVCAN_STM32_TIMER_NUMBER=${UAVCAN_STM32_TIMER_NUMBER}
```
pavel-kirienko: Ditto.

libuavcan_drivers/stm32/driver/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+include_directories(
+    ./src
```
pavel-kirienko: This path is not needed.

pavel-kirienko:
I still don't understand why we can't keep toolchain files in the application's project.

CMakeLists.txt
jgoppert: you can just list a bunch of paths here for each operating system, it is so you can do unit testing without having to include a nuttx build, do you want to make per operating system paths?

jgoppert:
I think you need some nuttx toolchain file for unit testing. It will be difficult for you to avoid breaking the px4 build otherwise. This was the most stripped down and self contained version I could make. If you can build against it, px4 should be fine.

libuavcan_drivers/stm32/driver/CMakeLists.txt
jgoppert: Removed.

CMakeLists.txt
jgoppert: I believe exposing the options to cmake/ the cmake-gui is important for end users. The typical way is to then use the cmake flags to drive a config.h.in -> config.h file using cmakedefine. If you note, when running the cmake-gui, this presents the user with a nice dialog with check boxes, and drop down lists for options.

LorenzMeier:
Pavel, did you have an opportunity to test this yet?

CMakeLists.txt
pavel-kirienko: The question is, can we tolerate extra maintenance requirements for the
sake of marginal improvement of user experience. Keeping in mind that all
of the relevant build config options are already documented here
http://uavcan.org/Implementations/Libuavcan/Tutorials/1._Library_build_configuration/,
I'm inclined to say that no, we can't.

pavel-kirienko:
I don't think we can be reasonably sure that px4 build is not broken
without actually building it. Since validation requires px4 anyway, there
seems to be no reason to keep toolchain files in this repo?

jgoppert:
The toolchain file was the only way I could debug the build system rapidly without having to commit, then pull on the px4 side, which was slow. To do a smoke test I was running these commands. I think it will reduce maintenance for us if you include this in your travis ci testing.

```bash
mkdir build
cd build
cmake .. -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchain-stm32-cortex-m4.cmake
make
```

If you feel strongly that this is a bad thing, maybe we should just move the driver code to the px4 repository, if that is where it is going to be tested anyways. I just don't want code tested in repo A, which is a part of repo B.

pavel-kirienko:
Seems like I'm missing something. Why did you have to commit&pull? You
could edit the libuavcan source directly in the px4's git submodule,
possibly having auto-update disabled
<https://github.com/PX4/Firmware/blob/master/Tools/check_submodules.sh#L3-L7>.
Since PX4's CI already builds the project with libuavcan included, I don't
see any extra maintenance effort here.

The current toolchain file not only requires maintenance, it is also too
specific for the PX4 project (assumes NuttX, STM32, Cortex M4F, and certain
PX4-specific set of compile flags).

> maybe we should just move the driver code to the px4 repository, if that
is where it is going to be tested anyways

Do you mean the STM32 platform driver? It is generic and it is used in many
other places, not only in PX4, so there's no reason to take it away.

jgoppert:
@pavel-kirienko  The goal at this point is just to get the new build tested with hardware, can you do that? Commands are below:

```bash
git clone git@github.com:jgoppert/Firmware
cd Firmware
git checkout cmake-uavcan
make px4fmu-v2_default
make px4fmu-v2_default upload
```

You can modify my PR as you wish, just don't break this externalproject_add call below:

(https://github.com/jgoppert/Firmware/blob/cmake-uavcan/src/modules/uavcan/CMakeLists.txt) 

```cmake
# uavcan project
set(uavcan_c_flags ${c_flags})
list(REMOVE_ITEM uavcan_c_flags -std=gnu++0x -D__CUSTOM_FILE_IO__)
set(uavcan_cxx_flags ${cxx_flags})
list(REMOVE_ITEM uavcan_cxx_flags -std=gnu++0x -std=c++11 -D__CUSTOM_FILE_IO__)
px4_join(OUT uavcan_c_flags LIST "${uavcan_c_flags}" GLUE " ")
px4_join(OUT uavcan_cxx_flags LIST "${uavcan_cxx_flags}" GLUE " ")
set(uavcan_deps git_uavcan)
set(uavcan_platform generic)
set(uavcan_definitions)

if (${OS} STREQUAL "nuttx")
	set(uavcan_platform stm32)
	list(APPEND uavcan_deps nuttx_export_${BOARD})
	list(APPEND uavcan_definitions
		-DUAVCAN_NO_ASSERTIONS=ON
		-DNUTTX_EXPORT_DIR=${CMAKE_BINARY_DIR}/${BOARD}/NuttX/nuttx-export
		-DUAVCAN_STM32_NUM_IFACES=2
		-DUAVCAN_USE_EXTERNAL_SNPRINTF=ON
		-DUAVCAN_USE_CPP03=ON
		-DUAVCAN_MEM_POOL_BLOCK_SIZE=48
		-DUAVCAN_MAX_NETWORK_SIZE_HINT=16
		-DUAVCAN_STM32_TIMER_NUMBER=5
		)
endif()
externalproject_add(libuavcan
	DEPENDS ${uavcan_deps}
	UPDATE_COMMAND git submodule update --init
	SOURCE_DIR ${CMAKE_SOURCE_DIR}/src/lib/uavcan
	CMAKE_ARGS -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
		-DUAVCAN_PLATFORM=${uavcan_platform}
		-DUAVCAN_OS=${OS}
		-DCMAKE_CXX_FLAGS=${uavcan_cxx_flags}
		-DCMAKE_C_FLAGS=${uavcan_c_flags}
		-DCMAKE_INSTALL_PREFIX=${ep_base}/Install
		${uavcan_definitions}
)

string(TOUPPER ${OS} OS_UPPER)

add_definitions(
	-DUAVCAN_NO_ASSERTIONS
	-DNUTTX_EXPORT_DIR=${CMAKE_BINARY_DIR}/${BOARD}/NuttX/nuttx-export
	-DUAVCAN_STM32_NUM_IFACES=2
	-DUAVCAN_USE_EXTERNAL_SNPRINT
	-DUAVCAN_MEM_POOL_BLOCK_SIZE=48
	-DUAVCAN_MAX_NETWORK_SIZE_HINT=16
	-DUAVCAN_STM32_TIMER_NUMBER=5
	-DUAVCAN_STM32_${OS_UPPER}=1
	-DUAVCAN_CPP_VERSION=UAVCAN_CPP03
	)

px4_add_module(
	MODULE modules__uavcan
	MAIN uavcan
	STACK 3200
	COMPILE_FLAGS
		-O3
	SRCS

		# Main
		uavcan_main.cpp
		uavcan_servers.cpp
		uavcan_params.c

		# Actuators
		actuators/esc.cpp

		# Sensors
		sensors/sensor_bridge.cpp
		sensors/gnss.cpp
		sensors/mag.cpp
		sensors/baro.cpp

	DEPENDS
		platforms__common
		libuavcan	
	)
```

So you can see that I'm not using the toolchain file that I embedded in uavcan, and can be removed if you wish.

pavel-kirienko:
> The goal at this point is just to get the new build tested with hardware, can you do that?

Yes it works. Tested with Zubax GNSS with dynamic node ID allocation and sensor data.

> So you can see that I'm not using the toolchain file that I embedded in uavcan, and can be removed if you wish.

Got it. I will have another look at your PR tomorrow; meanwhile I suggest to remove the build options as discussed, unless there are other objections.

pavel-kirienko:
Almost there! Now we need to figure out a way how to avoid [always including POSIX drivers into STM32 build](https://github.com/jgoppert/libuavcan/blob/cmake-cross/CMakeLists.txt#L79), and possibly also how to enable them when building for a generic platform (may come in handy for Windows builds, for example).

jgoppert:
Do we want to add back the OS flag then? We could try to use regex on the CMAKE_CXX_FLAGS if you want to avoid using cmake based config settings.

PR #77, opened 2015-10-25
Adding CONTINUOUS_INTEGRATION_BUILD flag

bendyer:
Opening PR to track discussion; I'll update this with the timing-sensitive tests prior to merge.

### DIFF
--- a/libuavcan/include/uavcan/build_config.hpp
+++ b/libuavcan/include/uavcan/build_config.hpp
@@ -151,13 +151,6 @@
 # define UAVCAN_USE_EXTERNAL_FLOAT16_CONVERSION 0
 #endif
 
-/*
- * Disables timing-dependent tests in the test suite, since these may fail on virtualized machines
- */
-#ifndef UAVCAN_NO_TIMING_SENSITIVE_TESTS
-# define UAVCAN_NO_TIMING_SENSITIVE_TESTS 0
-#endif
-
 /**
  * Run time checks.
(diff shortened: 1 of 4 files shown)


libuavcan/include/uavcan/build_config.hpp
```diff
@@ -151,6 +151,13 @@
 # define UAVCAN_USE_EXTERNAL_FLOAT16_CONVERSION 0
 #endif
 
+/*
+ * Disables timing-dependent tests in the test suite, since these may fail on virtualized machines
+ */
+#ifndef UAVCAN_NO_TIMING_SENSITIVE_TESTS
+# define UAVCAN_NO_TIMING_SENSITIVE_TESTS 0
```
pavel-kirienko: This option is development-specific, so it's better to keep it away from `build_config.hpp`. How about this solution: http://stackoverflow.com/a/7208148/1007777 - it is not intrusive to the library. We could also apply some naming pattern for time-sensitive tests, e.g. suffix them with `RealTime`, and then disable with `--gtest_filter=-*RealTime`.

bendyer:
I have re-run build 25.1 (https://travis-ci.org/UAVCAN/UAVCAN/libuavcan/jobs/87280203) eight times over the last 24 hours, but no failures yet. @pavel-kirienko, please could you merge this so that when the tests do finally fail again, we can identify them in the logs? The Travis UI won't show the full build log output, but they have a log download option.

PR #81, opened 2015-12-11
Returning meaningful error codes from STM32 CAN driver methods

pavel-kirienko:
instead of plain -1


pavel-kirienko:
Forcing CI

magicrub:
Why not use an enum?
Why return negative const value instead of the const value?

pavel-kirienko:
Casting error values to int removes enum type information, which is unsafe
and is explicitly prohibited by some coding standards (e.g. MISRA). This is
not what enums are for.

As for negation, it is a rather common practice. For instance, errno values
are handled this way in Linux kernel, also the rest of the library uses the
same convention.

PR #82, opened 2015-12-19
Can_acceptance_filter_configurator and unit test are corrected

ilia-sheremet:


libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -16,24 +16,22 @@ const unsigned CanAcceptanceFilterConfigurator::DefaultAnonMsgID;
 
 int16_t CanAcceptanceFilterConfigurator::loadInputConfiguration(AnonymousMessages load_mode)
 {
-    multiset_configs_.clear();
```
pavel-kirienko: We need `clear()` here.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -115,12 +103,27 @@ int16_t CanAcceptanceFilterConfigurator::computeConfiguration()
...
+    if (acceptance_filters_number == 0)
+    {
+        UAVCAN_TRACE("CanAcceptanceFilter", "No HW filters available");
+        return -ErrDriver;
+    }
+
+    if (filter_array_size > acceptance_filters_number)
+    {
+        UAVCAN_TRACE("CanAcceptanceFilter", "Too many filter configurations. Executing computeConfiguration()");
+        computeConfiguration(IgnoreAnonymousMessages);
```
pavel-kirienko: If I understand the logic correctly, this line is supposed to call `mergeConfigurations()`, not `computeConfiguration()`.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -115,12 +103,27 @@ int16_t CanAcceptanceFilterConfigurator::computeConfiguration()
...
+        filter_array_size = multiset_configs_.getSize();
+    }
 
     if (filter_array_size > MaxCanAcceptanceFilters)
     {
         UAVCAN_ASSERT(0);
         return -ErrLogic;
+        UAVCAN_TRACE("CanAcceptanceFilter", "Failed to apply HW filter configuration");
```
pavel-kirienko: Dead code.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -137,19 +150,21 @@ int16_t CanAcceptanceFilterConfigurator::applyConfiguration(void)
         if (iface == NULL)
         {
             return -ErrDriver;
+            UAVCAN_TRACE("CanAcceptanceFilter", "Failed to apply HW filter configuration");
```
pavel-kirienko: Dead code again. I will ignore other instances of this mistake.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
ilia-sheremet: But I delete it  in case the addFilterConfig() was called before computeConfiguration(). Or/and if configs number > available HW  when you call applyConfiguration, than the computeConfiguration called, thus loadInputConfiguration

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -182,8 +191,10 @@ int CanAcceptanceFilterConfigurator::configureFilters(AnonymousMessages mode)
 
 uint16_t CanAcceptanceFilterConfigurator::getNumFilters() const
 {
+	if (filters_number_ == 0)
+	{
     static const uint16_t InvalidOut = 0xFFFF;
-    uint16_t out = InvalidOut;
+        uint16_t out = InvalidOut;
```
pavel-kirienko: Formatting. I suggest to reformat all of the changed files with uncrustify.

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
pavel-kirienko: This logic is highly complicated. One would expect that the result of `loadInputConfiguration()` would be having the config multiset loaded with unoptimized set of filters. The fact that the multiset is not getting cleared here is highly counter-intuitive. As for `applyConfiguration()`, see my comment there.

ilia-sheremet: Agree, so I will mention in tutorial and header that computeConfiguration() must be called before addFilterConfig().

libuavcan/src/transport/uc_can_acceptance_filter_configurator.cpp
```diff
@@ -204,6 +215,22 @@ uint16_t CanAcceptanceFilterConfigurator::getNumFilters() const
...
+	}
+	else
+	{
+		return filters_number_;
+	}
+}
+
+int16_t CanAcceptanceFilterConfigurator::addFilterConfig(const CanFilterConfig& config)
+{
+    CanFilterConfig supl_config = config;
```
pavel-kirienko: I guess without this, the compiler was complaining about the reference being const? The proper fix is as follows: `multiset_configs_.emplace<const CanFilterConfig&>(config)`.

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
```diff
@@ -148,6 +148,9 @@ class SocketCanIface : public uavcan::ICanIface
     std::queue<RxItem> rx_queue_;                                   // TODO: Use pool allocator
     std::unordered_multiset<std::uint32_t> pending_loopback_ids_;   // TODO: Use pool allocator
 
+    std::vector<uavcan::CanFilterConfig> hw_filters_container_;
+    bool filters_configured_ = false;
```
pavel-kirienko: This variable can be replaced with a check like `!hw_filters_container_.empty()`. I suggest to remove it. It's always better to have as few states as possible.

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
```diff
@@ -301,7 +304,17 @@ class SocketCanIface : public uavcan::ICanIface
...
+                    {
+                        if (checkHWFilters(rx.frame))
+                        {
+                            rx_queue_.push(rx);
+                        }
+                    }
+                    else
+                    {
+                        rx_queue_.push(rx);
+                    }
```
pavel-kirienko: Here, the check on whether the filter array is empty or not should be delegated to `checkHWFilters()` entirely. Therefore, the code should be like this:

```c++
if (checkHWFilters(rx.frame))
{
    rx_queue_.push(rx);
}
```

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
```diff
@@ -315,6 +328,22 @@ class SocketCanIface : public uavcan::ICanIface
...
+    {
+        uint16_t container_size = hw_filters_container_.size();
+        for (uint16_t i = 0; i < container_size; i++)
+        {
+            if (((frame.id & hw_filters_container_[i].mask) ^ hw_filters_container_[i].id) == 0)
+            {
+                return 1;
+            }
+        }
+        return 0;
```
pavel-kirienko: You need to use explicit true/false with boolean data type. Implicitly casting int to bool is unsafe, and many code analysis tools will rightfully complain about it.

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
```diff
@@ -404,38 +433,46 @@ class SocketCanIface : public uavcan::ICanIface
...
-        int ret = setsockopt(fd_, SOL_CAN_RAW, CAN_RAW_FILTER, filts.data(), sizeof(::can_filter) * num_configs);
-        return (ret < 0) ? -1 : 0;
+
+        if (! hw_filters_container_.empty())
+        {
+            filters_configured_ = true;
+            return 0;
+        }
+
+        return -1;
```
pavel-kirienko: `num_configs` set to zero should not be considered error.

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
```diff
@@ -404,38 +433,46 @@ class SocketCanIface : public uavcan::ICanIface
...
     }
 
     /**
      * SocketCAN emulates the CAN filters in software, so the number of filters is virtually unlimited.
      * This method returns a constant value.
      */
-    std::uint16_t getNumFilters() const override { return 255; }
+    std::uint16_t getNumFilters() const override { return 8; }
```
pavel-kirienko: We need a static constexpr member constant for this.

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
```diff
@@ -404,38 +433,46 @@ class SocketCanIface : public uavcan::ICanIface
...
             }
             if (fc.mask & uavcan::CanFrame::FlagEFF)
             {
-                filts[i].can_mask |= CAN_EFF_FLAG;
+                hw_filters_container_[i].mask |= CAN_EFF_FLAG;
             }
             if (fc.mask & uavcan::CanFrame::FlagRTR)
             {
-                filts[i].can_mask |= CAN_RTR_FLAG;
+                hw_filters_container_[i].mask |= CAN_RTR_FLAG;
```
pavel-kirienko: You are mixing types from libuavcan and from SocketCAN, this is dangerous. The container `hw_filters_container_` contains objects of type `uavcan::CanFilterConfig`, you can't use SocketCAN constants with them; instead, you should use this: https://github.com/ilia-sheremet/uavcan/blob/master/libuavcan/include/uavcan/driver/can.hpp#L26-L30.

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
pavel-kirienko: Also, loopback frames must be accepted unconditionally, otherwise we would be having the same problem we had with SocketCAN filters.

pavel-kirienko:
Also it is preferable to move the filter checks into the [`read()` method](https://github.com/ilia-sheremet/uavcan/blob/master/libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp#L206). This way we'd be able to save processing time on converting frames from SocketCAN to libuavcan representation.

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
```diff
@@ -225,6 +227,16 @@ class SocketCanIface : public uavcan::ICanIface
         {
             return (res < 0 && errno == EWOULDBLOCK) ? 0 : res;
         }
+        /*
+         * Flags
+         */
+        loopback = (msg.msg_flags & static_cast<int>(MSG_CONFIRM)) != 0;
+
+        if (!checkHWFilters(sockcan_frame) && !loopback)
```
pavel-kirienko: Order should be reversed: `if (!loopback && !checkHWFilters(sockcan_frame))`, because there is no point in checking HW filters on loopback frames.

libuavcan_drivers/linux/include/uavcan_linux/socketcan.hpp
```diff
@@ -315,6 +323,29 @@ class SocketCanIface : public uavcan::ICanIface
         }
     }
 
+    /**
+     * Returns true if a frame accepted by HW filters
+     */
+    bool checkHWFilters(const ::can_frame frame) const
```
pavel-kirienko: Pass by reference.

pavel-kirienko:
:+1:

PR #85, opened 2016-02-15
support for FreeRTOS

suiauthon:
added support for FreeRTOS. New macro are added (UAVCAN_SMT32_FREERTOS).

### DIFF
--- a/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
+++ b/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
@@ -168,7 +168,7 @@ public:
 class BusEvent
 {
     SemaphoreHandle_t sem_;
-    BaseType_t xHigherPriorityTaskWoken;
+    BaseType_t higher_priority_task_woken;
 
 public:
     BusEvent(CanDriver& can_driver)
--- a/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
+++ b/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
@@ -18,7 +18,6 @@
 # include <cstring>
 #elif UAVCAN_STM32_BAREMETAL
 #elif UAVCAN_STM32_FREERTOS
-# include <stm32f4xx.h>
 # include <cmsis_os.h>
 #else
 # error "Unknown OS"
--- a/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
+++ b/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
@@ -17,6 +17,12 @@
 # include <ctime>
 # include <cstring>
 #elif UAVCAN_STM32_BAREMETAL
+#elif UAVCAN_STM32_FREERTOS
+# include <stm32f4xx.h>
+# include <cmsis_os.h>
+  #ifndef MAX_SEMAPHORE_COUNT
+    #define MAX_SEMAPHORE_COUNT      50
+  #endif
 #else
 # error "Unknown OS"
 #endif
@@ -136,7 +142,7 @@ public:
         (void)can_driver;
     }
 
-    bool wait(uavcan::MonotonicDuration)
+    bool wait(uavcan::MonotonicDuration duration)
     {
         (void)duration;
         bool lready = ready;
@@ -159,6 +165,40 @@ class Mutex
 public:
     void lock() { }
     void unlock() { }
+};
+
+#elif UAVCAN_STM32_FREERTOS
+
+class BusEvent
+{
+    SemaphoreHandle_t sem_;
+    BaseType_t xHigherPriorityTaskWoken;
+
+public:
+    BusEvent(CanDriver& can_driver)
+    {
+        (void)can_driver;
+        sem_ = xSemaphoreCreateCounting( MAX_SEMAPHORE_COUNT, 0 );
+    }
+
+    bool wait(uavcan::MonotonicDuration duration);
+
+    void signal();
+
+    void signalFromInterrupt();
+};
+
+class Mutex
+{
+    SemaphoreHandle_t mtx_;
+    BaseType_t xHigherPriorityTaskWoken;
+public:
+    Mutex(void)
+    {
+        mtx_ = xSemaphoreCreateMutex();
+    }
+    void lock();
+    void unlock();
 };
 
 #endif
--- a/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
+++ b/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
@@ -20,9 +20,6 @@
 #elif UAVCAN_STM32_FREERTOS
 # include <stm32f4xx.h>
 # include <cmsis_os.h>
-#   ifndef MAX_SEMAPHORE_COUNT
-#   define MAX_SEMAPHORE_COUNT      50
-#endif
 #else
 # error "Unknown OS"
 #endif
@@ -178,7 +175,7 @@ public:
     BusEvent(CanDriver& can_driver)
     {
         (void)can_driver;
-        sem_ = xSemaphoreCreateCounting( MAX_SEMAPHORE_COUNT, 0 );
+        sem_ = xSemaphoreCreateBinary();
     }
 
     bool wait(uavcan::MonotonicDuration duration);
--- a/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
+++ b/libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
@@ -20,9 +20,9 @@
 #elif UAVCAN_STM32_FREERTOS
 # include <stm32f4xx.h>
 # include <cmsis_os.h>
-  #ifndef MAX_SEMAPHORE_COUNT
-    #define MAX_SEMAPHORE_COUNT      50
-  #endif
+#   ifndef MAX_SEMAPHORE_COUNT
+#   define MAX_SEMAPHORE_COUNT      50
+#endif
 #else
 # error "Unknown OS"
 #endif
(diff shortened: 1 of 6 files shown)


pavel-kirienko:
Thanks. I noticed that you also changed baremetal support. Is it fixed now? Have you had a chance to test it?

libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
```diff
@@ -17,6 +17,12 @@
 # include <ctime>
 # include <cstring>
 #elif UAVCAN_STM32_BAREMETAL
+#elif UAVCAN_STM32_FREERTOS
+# include <stm32f4xx.h>
+# include <cmsis_os.h>
+  #ifndef MAX_SEMAPHORE_COUNT
+    #define MAX_SEMAPHORE_COUNT      50
+  #endif
```
pavel-kirienko: Please fix formatting above: the octothorpe symbol (`#`) should be placed on the beginning of the line.

libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
```diff
@@ -161,6 +167,40 @@ class Mutex
...
+
+    void signal();
+
+    void signalFromInterrupt();
+};
+
+class Mutex
+{
+    SemaphoreHandle_t mtx_;
+    BaseType_t xHigherPriorityTaskWoken;
```
pavel-kirienko: This field is not used.

libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
```diff
@@ -161,6 +167,40 @@ class Mutex
     void unlock() { }
 };
 
+#elif UAVCAN_STM32_FREERTOS
+
+class BusEvent
+{
+    SemaphoreHandle_t sem_;
+    BaseType_t xHigherPriorityTaskWoken;
```
pavel-kirienko: This doesn't have to be a class field, make it local variable.

libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
```diff
@@ -17,6 +17,12 @@
 # include <ctime>
 # include <cstring>
 #elif UAVCAN_STM32_BAREMETAL
+#elif UAVCAN_STM32_FREERTOS
+# include <stm32f4xx.h>
+# include <cmsis_os.h>
+  #ifndef MAX_SEMAPHORE_COUNT
+    #define MAX_SEMAPHORE_COUNT      50
```
pavel-kirienko: Why 50? What does it mean?

libuavcan_drivers/stm32/driver/src/internal.hpp
```diff
@@ -122,6 +134,22 @@ struct CriticalSectionLocker
     }
 };
 
+#elif UAVCAN_STM32_FREERTOS
+
+struct CriticalSectionLocker
+{
+    //TODO napisati funkcije
```
pavel-kirienko: English please. ;)

What function do you want to write here? Everything seems to be in place, am I missing something?

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -472,6 +477,9 @@ bool CanIface::waitMsrINakBitStateChange(bool target_state)
 #if UAVCAN_STM32_CHIBIOS
         ::chThdSleep(MS2ST(1));
 #endif
+#if UAVCAN_STM32_FREERTOS
+        //::osDelay(1);
```
pavel-kirienko: Some sort of rescheduling delay is necessary here. Probably you want to uncomment this call.

libuavcan_drivers/stm32/driver/src/uc_stm32_thread.cpp
```diff
@@ -79,6 +80,56 @@ void Mutex::unlock()
...
+    xSemaphoreGive( sem_ );
+}
+
+void BusEvent::signalFromInterrupt()
+{
+    xHigherPriorityTaskWoken = pdFALSE;
+
+    xSemaphoreGiveFromISR( sem_, &xHigherPriorityTaskWoken );
+    
+    portYIELD_FROM_ISR( xHigherPriorityTaskWoken );
```
pavel-kirienko: This seems to be a mis-use of the FreeRTOS API. Correct me if I'm wrong, but `portYIELD_FROM_ISR()` must be invoked only at the end of the ISR, which is not the case here - the driver may continue to perform other tasks in the ISR even after `signalFromInterrupt()` is invoked.

libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
suiauthon: MAX_SEMAPHORE_COUNT is the maximum count value of the counting semaphore that can be reached. When the semaphore reaches this value it can no longer be 'given'. I assumes that 50 will be enough, if am wrong please tell me how much should i put here.

pavel-kirienko: This is the only place where the semaphore count can be decremented: https://github.com/larics/libuavcan/blob/master/libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp#L874. As you can see, it is not guaranteed to be reachable. Therefore, the semaphore may overflow. I suggest to use a different synchronization primitive, probably an idempotent one, like a conditional variable or whatever FreeRTOS has to offer nowadays.

libuavcan_drivers/stm32/driver/src/internal.hpp
```diff
@@ -14,7 +14,10 @@
 # include <chip/stm32_tim.h>
 # include <syslog.h>
 #elif UAVCAN_STM32_BAREMETAL
-# include <chip.h>
+# include <stm32f4xx.h>
```
nuft: The driver should support all STM32 versions.

pavel-kirienko:
@suiauthon any update on this?

libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
suiauthon: Sorry but i don't have solution for this.

pavel-kirienko: Okay, can we use a binary semaphore here? See http://www.freertos.org/Embedded-RTOS-Binary-Semaphores.html

pavel-kirienko:
@suiauthon Do you need any help with this?

suiauthon:
@pavel-kirienko  I think that is ready to be merged.

pavel-kirienko:
There's still one issue with a class field that should be converted into a local variable, see my comment above.

libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
suiauthon: I think that should stay class variable because i'm using it in function yieldFromISR. @pavel-kirienko

libuavcan_drivers/stm32/driver/include/uavcan_stm32/thread.hpp
```diff
@@ -17,6 +17,9 @@
 # include <ctime>
 # include <cstring>
 #elif UAVCAN_STM32_BAREMETAL
+#elif UAVCAN_STM32_FREERTOS
+# include <stm32f4xx.h>
```
pavel-kirienko: Can we replace this with a more generic include that doesn't depend on a particular MCU model?

libuavcan_drivers/stm32/driver/src/internal.hpp
```diff
@@ -15,6 +15,9 @@
 # include <syslog.h>
 #elif UAVCAN_STM32_BAREMETAL
 # include <chip.h>
+#elif UAVCAN_STM32_FREERTOS
+# include <stm32f4xx.h>
```
pavel-kirienko: Same here

libuavcan_drivers/stm32/driver/src/internal.hpp
```diff
@@ -122,6 +134,22 @@ struct CriticalSectionLocker
...
+{
+
+    CriticalSectionLocker()
+    {
+      taskENTER_CRITICAL();
+    }
+
+    ~CriticalSectionLocker()
+    {
+      taskEXIT_CRITICAL();
```
pavel-kirienko: Please use 4 spaces per indent

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -16,6 +16,9 @@
 # include <arch/board/board.h>
 #elif UAVCAN_STM32_BAREMETAL
 #include <chip.h>
+#elif UAVCAN_STM32_FREERTOS
+#include <stm32f4xx.h>
+#include <cmsis_os.h>
```
pavel-kirienko: These includes are redundant, see `internal.hpp` (yes, some of the includes above them are also redundant, that's a separate problem).

pavel-kirienko:
Could you please reformat the code to match [the style guide](https://github.com/Zubax/zubax_style_guide). At least the following things are not compliant:

- No spaces inside braces: `(xxx)`
- Spaces around operators: `x / y`
- Indentation is 4 spaces per level

Also, once we're finished with this PR, could you please draft a few sentences describing the configuration options pertaining to FreeRTOS support here: http://uavcan.org/Implementations/Libuavcan/Platforms/STM32/ ? This page is hosted at https://github.com/UAVCAN/uavcan.github.io/blob/master/Implementations/Libuavcan/Platforms/STM32/index.md. Thanks in advance!

suiauthon:
@pavel-kirienko I think that's it.

pavel-kirienko:
Thanks Marko, great job :+1:

pavel-kirienko:
Hey @suiauthon, I mentioned FreeRTOS support on the driver docs page here http://uavcan.org/Implementations/Libuavcan/Platforms/STM32/. Do you think there is something else that needs to be said about your code, particularly about its usage?

suiauthon:
That's all. Only UAVCAN_STM32_FREERTOS should be set to 1.

PR #90, opened 2016-04-17
cmake only run libuavcan_dsdlc if there are changes

dagar:
Right now every single incremental PX4 build reruns dsdl_compiler/libuavcan_dsdlc.


LorenzMeier:
@dagar Looking nice! but the build fails.

dagar:
I fixed the build and hid some of the configure output behind UAVCAN_CMAKE_VERBOSE. It was somewhat confusing because the flags are printed at the beginning of the PX4 build, but they aren't necessarily the same flags used outside of UAVCAN.

PR #92, opened 2016-04-27
Do Not merge - This is needed to work with the latest upstream nuttx

davids5:
@pavel-kirienko 

This is just a second place holder on top of the latest version PX4 is using of libuavcan on their master to build with upstream nuttx


davids5:
@pavel-kirienko - this can now go into master.

PR #98, opened 2017-01-12
For upstream nuttx changes 4 fixed CI

davids5:
@pavel-kirienko - This was the rest of that changes to support the upstream nuttx. Once this passed CI - I will rebase it on master and then have it tested with PX4.


davids5:
This looks good to go now. I do not see a need to rebase once merged to master I will update PX4 to the current master master and  have it tested with PX4.

PR #105, opened 2017-05-06
stm32 driver: filter initialization correction

EShamaev:
This seems to be a severe bug:

- FMR register with settings of 28 will not allow any reception on CAN2
- FilterRegister[NumFilters] is out of bound
- FA1R = 1 | (1 << NumFilters) has bit number 28 set, which has no sense


EShamaev:
Sorry, in my version NumFilters was 28

Issue #58, opened 2015-08-16
STM32 acceptance filters


pavel-kirienko:
The `configureFilters()` method should be implemented. https://github.com/UAVCAN/libuavcan/blob/master/libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp#L450-L458

EShamaev:
https://github.com/UAVCAN/libuavcan/pull/106

PR #106, opened 2017-05-06
stm32 driver configuration of filters

EShamaev:
According to https://github.com/UAVCAN/libuavcan/issues/58

### DIFF
--- a/libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
+++ b/libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
@@ -450,11 +450,84 @@ uavcan::int16_t CanIface::receive(uavcan::CanFrame& out_frame, uavcan::Monotonic
 uavcan::int16_t CanIface::configureFilters(const uavcan::CanFilterConfig* filter_configs,
                                            uavcan::uint16_t num_configs)
 {
-    // TODO: Hardware filter support
-    CriticalSectionLocker lock;
-    (void)filter_configs;
-    (void)num_configs;
-    return -ErrNotImplemented;
+    if (num_configs <= NumFilters)
+    {
+        CriticalSectionLocker lock;
+
+        can_->FMR |= bxcan::FMR_FINIT;
+
+        // Slave (CAN2) gets half of the filters
+        can_->FMR &= ~0x00003F00UL;
+        can_->FMR |= static_cast<uint32_t>(NumFilters) << 8;
+
+        can_->FFA1R = 0x0AAAAAAA; // FIFO's are interleaved between filters
+        can_->FM1R = 0; // Identifier Mask mode
+        can_->FS1R = 0x7ffffff; // Single 32-bit for all
+
+        const uint8_t filter_start_index = (self_index_ == 0) ? 0 : NumFilters;
+
+        if (num_configs == 0)
+        {
+            can_->FilterRegister[filter_start_index].FR1 = 0;
+            can_->FilterRegister[filter_start_index].FR2 = 0;
+            can_->FA1R = 1 << filter_start_index;
+        }
+        else
+        {
+            for (uint8_t i = 0; i < NumFilters; i++)
+            {
+                if (i < num_configs)
+                {
+                    uint32_t id   = 0;
+                    uint32_t mask = 0;
+
+                    const uavcan::CanFilterConfig* const cfg = filter_configs + i;
+
+                    if ((cfg->id & uavcan::CanFrame::FlagEFF) || !(cfg->mask & uavcan::CanFrame::FlagEFF))
+                    {
+                        id   = (cfg->id   & uavcan::CanFrame::MaskExtID) << 3;
+                        mask = (cfg->mask & uavcan::CanFrame::MaskExtID) << 3;
+                        id |= bxcan::RIR_IDE;
+                    }
+                    else
+                    {
+                        id   = (cfg->id   & uavcan::CanFrame::MaskStdID) << 21;  // Regular std frames, nothing fancy.
+                        mask = (cfg->mask & uavcan::CanFrame::MaskStdID) << 21;  // Boring.
+                    }
+
+                    if (cfg->id & uavcan::CanFrame::FlagRTR)
+                    {
+                        id |= bxcan::RIR_RTR;
+                    }
+
+                    if (cfg->mask & uavcan::CanFrame::FlagEFF)
+                    {
+                        mask |= bxcan::RIR_IDE;
+                    }
+
+                    if (cfg->mask & uavcan::CanFrame::FlagRTR)
+                    {
+                        mask |= bxcan::RIR_RTR;
+                    }
+
+                    can_->FilterRegister[filter_start_index + i].FR1 = id;
+                    can_->FilterRegister[filter_start_index + i].FR2 = mask;
+
+                    can_->FA1R |= (1 << (filter_start_index + i));
+                }
+                else
+                {
+                    can_->FA1R &= ~(1 << (filter_start_index + i));
+                }
+            }
+        }
+
+        can_->FMR &= ~bxcan::FMR_FINIT;
+
+        return 0;
+    }
+
+    return -ErrFilterNumConfigs;
 }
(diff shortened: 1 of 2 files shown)


libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -450,10 +450,35 @@ uavcan::int16_t CanIface::receive(uavcan::CanFrame& out_frame, uavcan::Monotonic
...
-    // TODO: Hardware filter support
-    CriticalSectionLocker lock;
-    (void)filter_configs;
-    (void)num_configs;
+    if (num_configs < (NumFilters / 2)) {
+        CriticalSectionLocker lock;
+
+        can_->FMR |= bxcan::FMR_FINIT;
+
+        can_->FMR = (can_->FMR & ~0x00003F00) | static_cast<uint32_t>(NumFilters / 2) << 8; // Slave (CAN2) gets half of the filters
```
pavel-kirienko: Line is too long

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -450,10 +450,35 @@ uavcan::int16_t CanIface::receive(uavcan::CanFrame& out_frame, uavcan::Monotonic
 uavcan::int16_t CanIface::configureFilters(const uavcan::CanFilterConfig* filter_configs,
                                            uavcan::uint16_t num_configs)
 {
-    // TODO: Hardware filter support
-    CriticalSectionLocker lock;
-    (void)filter_configs;
-    (void)num_configs;
+    if (num_configs < (NumFilters / 2)) {
```
pavel-kirienko: Braces should be put after a new line.

pavel-kirienko: `NumFilters` is set to 14, which is already half of the amount of filters available on dual-CAN devices. Therefore I suggest to remove the division by two here and below.

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -450,10 +450,35 @@ uavcan::int16_t CanIface::receive(uavcan::CanFrame& out_frame, uavcan::Monotonic
...
+        can_->FFA1R = 0x0AAAAAAA; // FIFO's are interleaved between filters
+        can_->FM1R = 0; // Identifier Mask mode
+        can_->FS1R = 0x7ffffff; // Single 32-bit for all
+
+        const uint8_t filter_start_index = (self_index_ == 0) ? 0 : NumFilters / 2;
+
+        for (uint8_t i = 0; i < (NumFilters / 2); i++) {
+            if (i < num_configs) {
+                can_->FilterRegister[filter_start_index + i].FR1 = filter_configs[i].id;
+                can_->FilterRegister[filter_start_index + i].FR2 = filter_configs[i].mask;
```
pavel-kirienko: The bit masks (EFF, RTR) should be converted from the UAVCAN format to the format used by bxCAN. There is also one horrible undocumented caveat I spent hours on, it is described in the STM32 driver from libcanard. Here: https://github.com/UAVCAN/libcanard/blob/master/drivers/stm32/canard_stm32.c#L449-L526

> If Mask asks us to accept both STDID and EXTID, we need to use EXT mode on the filter,
> otherwise it will reject all EXTID frames. This logic is not documented in the RM.

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -450,10 +450,35 @@ uavcan::int16_t CanIface::receive(uavcan::CanFrame& out_frame, uavcan::Monotonic
...
+                can_->FA1R &= ~(1 << (filter_start_index + i));
+            }
+        }
+
+        can_->FMR &= ~bxcan::FMR_FINIT;
+
+        return 0;
+    }
+
     return -ErrNotImplemented;
```
pavel-kirienko: Are you sure? ;)

EShamaev:
Sorry, in my version NumFilters was 28

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
EShamaev: I was aware of that, thought it was made in lib, not here. Will add.

EShamaev:
Actually, mapping of bits is documented in RM:
image

pavel-kirienko:
Sorry I wasn't referring to the mapping. What isn't documented (or I couldn't find where) is that in order to receive both STD and EXT frames, the EXT bit in the mask register needs to be set (IIRC).

EShamaev:
Now this looks good.

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -69,6 +69,8 @@ static int can2_irq(const int irq, void*);
 #define CAN1_RX1_IRQHandler    CAN_RX1_IRQHandler
 #endif
 
+#define STM32_CAN_RIR_RTR                          (1U << 1)            // Bit 1: Remote Transmission Request
+#define STM32_CAN_RIR_IDE                          (1U << 2)            // Bit 2: Identifier Extension
```
pavel-kirienko: Why do we need this here? These bits are already defined in bxcan.hpp, here: https://github.com/EShamaev/libuavcan/blob/f064efa80dba9ff125d7fded8c9f63bc3a4639bf/libuavcan_drivers/stm32/driver/include/uavcan_stm32/bxcan.hpp#L242-L243

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -450,11 +452,74 @@ uavcan::int16_t CanIface::receive(uavcan::CanFrame& out_frame, uavcan::Monotonic
...
+                if (cfg->mask & uavcan::CanFrame::FlagRTR)
+                {
+                    mask |= STM32_CAN_RIR_RTR;
+                }
+
+                can_->FilterRegister[filter_start_index + i].FR1 = id;
+                can_->FilterRegister[filter_start_index + i].FR2 = mask;
+
+                can_->FA1R |= (1 << (filter_start_index + i));
+            } else
```
pavel-kirienko: Ouch

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -450,11 +452,74 @@ uavcan::int16_t CanIface::receive(uavcan::CanFrame& out_frame, uavcan::Monotonic
...
-    (void)num_configs;
-    return -ErrNotImplemented;
+    if (num_configs < NumFilters)
+    {
+        CriticalSectionLocker lock;
+
+        can_->FMR |= bxcan::FMR_FINIT;
+
+        // Slave (CAN2) gets half of the filters
+        can_->FMR &= ~0x00003F00;
```
pavel-kirienko: You're applying ~ to a signed value; perhaps you meant to say `~0x00003F00UL`?

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
EShamaev: Did not find that, my mistake

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
EShamaev: I miss LINT...

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
EShamaev: ouch

pavel-kirienko:
My nitpicking skills are failing me. Are we ready to merge now?

EShamaev:
Added a major check. I believe it should be done.

PR #107, opened 2017-05-24
fixed usage of atomic operation lib for armcc compiler, added conditi…

suiauthon:
…onal compilation


pavel-kirienko:
Due to lack of response from @suiauthon, I'm assuming this to be ready to merge.

PR #119, opened 2017-11-29
Minor fixes: unnecessary sign conversion, redundant path delimiter.

battlesnake:
Build failed with -Wsign-conversion -Werror.  The offending conversion isn't necessary since it's testing equality to positive-one.

The redundant path delimiter is not a bug, more a minor style issue.  Feel free to ignore that commit.


pavel-kirienko:
:+1:

PR #122, opened 2018-03-25
Rename variable "ch" to "chr" to avoid shadow variable in ChibiOS

magicrub:
Rename variable "ch" to "chr" to avoid shadow variable in ChibiOS. There is a global variable in ChibiOS called "ch" so when a local variable in UAVCAN is called the same name there is a shadowing compile warning. It's never a real conflict but it causes a bunch of compiler warnings. Simply renaming the local variable solves it.


pavel-kirienko:
:+1:

PR #128, opened 2018-05-19
DSDL union tag code generation bug fix

pavel-kirienko:
Report from Juha Kuikka @kuikka on [Gitter](https://gitter.im/UAVCAN/general) and the [mailing list](https://groups.google.com/d/msgid/uavcan/4df274d9-1822-41e1-bd95-03f1ba30d007%40googlegroups.com?utm_medium=email&utm_source=footer).

> I'm seeing what looks like a bug with unions in messages. This issue is visible when I am trying to communicate between libuavcan and pyuavcan. I have an union with 4 members but the tag comes out to 3 bits by libuavcan due to `typedef ::uavcan::IntegerSpec< ::uavcan::IntegerBitLen< 4 >::Result, ::uavcan::SignednessUnsigned, ::uavcan::CastModeTruncate > TagType;` this is because `IntegerBitLen<4>::Result` evaluates to 3, which makes the tag size. The python side seems to set it to 2 bits because incoming data is off by a bit.

Concerning this:

> The data_type_template.tmpl could be changed have a -1 but that would break @union with only one member.

Here's a relavant quote from the specification:

> Unions are required to have at least two fields.

### DIFF
--- a/libuavcan/dsdl_compiler/libuavcan_dsdl_compiler/data_type_template.tmpl
+++ b/libuavcan/dsdl_compiler/libuavcan_dsdl_compiler/data_type_template.tmpl
@@ -70,7 +70,7 @@ struct UAVCAN_EXPORT ${t.cpp_type_name}
         };
     };
 
-    typedef ::uavcan::IntegerSpec< ::uavcan::IntegerBitLen< ${len(fields)} >::Result,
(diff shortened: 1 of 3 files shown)


kuikka:
Perfect! Fixes the problem.

PR #129, opened 2018-05-26
stm32: allow for less than 1ms wait time on ChibiOS

tridge:
this prevents us chewing all the cpu when asking for fast spin times


pavel-kirienko:
Thanks Andrew!

PR #130, opened 2018-06-03
uavcan_stm32: add apis to support ardupilot way of using uavcan

magicrub:
uavcan_stm32: add apis to support ardupilot way of using uavcan

These changes are to support ArduPilot hardware infrastructure differences over multiple embedded OSs including Nuttx, ChibiOS, BareMetal, and  FreeRTOS implementations.

### DIFF
--- a/libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
+++ b/libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
@@ -976,7 +976,7 @@ static void nvicEnableVector(IRQn_Type irq,  uint8_t prio)
 
 #endif
 
-void CanDriver::initOnce()
+void CanDriver::initOnce(uavcan::uint8_t can_number)
 {
     /*
      * CAN1, CAN2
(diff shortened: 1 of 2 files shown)


pavel-kirienko:
Thanks, Tom. As I said in the chat room, can we please find some way of reducing the amount of duplicated code?

magicrub:
well, I've reduced it so init() just loops through init(i) and it takes less LOC y only doing the actual work in a single function but the compiled output on ArduPilot is 16 bytes larger. So, it seems the compiler is better at optimizing than me/humans. However, its easier to maintain.

Let's not merge yet. I'd like to test a little more

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -1091,6 +1097,38 @@ int CanDriver::init(const uavcan::uint32_t bitrate, const CanIface::OperatingMod
     return res;
 }
 
+int CanDriver::configureIface(const uavcan::uint32_t bitrate, const CanIface::OperatingMode mode, uavcan::uint8_t can_number)
+{
+    int res = 0;
+
+    UAVCAN_STM32_LOG("Initing iface %i...", can_number);
+
+    if (can_number == 0) {
```
pavel-kirienko: Opening brace goes on the next line.

magicrub:
I'll test on ArduPilot, could someone else please test on other platforms?

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -1091,6 +1097,38 @@ int CanDriver::init(const uavcan::uint32_t bitrate, const CanIface::OperatingMod
...
+    }
+#if UAVCAN_STM32_NUM_IFACES > 1
+    else if (can_number == 1) {
+        /*
+        * CAN2
+        */
+        ifaces[1] = &if1_;                          // Same thing here.
+        res = if1_.init(bitrate, mode);
+    }
+#endif
```
pavel-kirienko: Per MISRA, every condition that contains an else-if should have a plain else clause at the end. In this case, it must handle the case when `can_number` is neither 0 nor 1. There are several other occurrences where this should be updated.

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
```diff
@@ -1042,44 +1058,34 @@ void CanDriver::initOnce()
...
-#if UAVCAN_STM32_NUM_IFACES > 1
-    UAVCAN_STM32_LOG("Initing iface 1...");
-    ifaces[1] = &if1_;                          // Same thing here.
-    res = if1_.init(bitrate, mode);
-    if (res < 0)
-    {
-        UAVCAN_STM32_LOG("Iface 1 init failed %i", res);
-        ifaces[1] = UAVCAN_NULLPTR;
+    res = configureIface(bitrate, mode, can_number);
+    if (res < 0) {
```
pavel-kirienko: **BRACE**
image

libuavcan_drivers/stm32/driver/src/uc_stm32_can.cpp
magicrub: how about a switch statement with an empty default?

pavel-kirienko: If you find that reasonable, sure (in my opinion `if` works well here)

magicrub: nevermind, it gets messy

magicrub:
all braces fixed and I've added error handling with an else in two places

pavel-kirienko:
Looking at the number of changes and considering the fact that they are temporary, I am beginning to question whether it all makes sense. Would it not be better to just fork the driver immediately and maintain your own copy?

magicrub:
it's been forked for a year. The change was small until you wanted it optimized. Optimizing it made it messy and made the program size larger.

pavel-kirienko:
You are right, I should have foreseen this earlier, my fault. I'm just saying that the ArduPilot project might benefit from forking the driver completely and removing all of the unnecessary stuff from it (like FreeRTOS and baremetal support). I am going to remove all of the hardware-specific stuff from this repo in the future anyway. Do you still want this PR merged?

magicrub:
I brought this up during the weekly ArduPilot Dev Call. We decided that we should merge this in (after more testing) and then we'll update the ArduPilot fork to match this upstream repo. 

Please hold the merge until we can test it a little more

pavel-kirienko:
Closing because the drivers are being moved out of the repo (https://github.com/UAVCAN/libuavcan/pull/143); if still relevant, please submit this changeset to the new repo at https://github.com/UAVCAN/libuavcan_stm32.

pavel-kirienko:
But I would suggest you instead to just create your own driver with your preferred API.

Issue #132, opened 2018-06-07
googletest is not an explicit dependency


thirtytwobits:
Googletest/googlemock shouldn't come from the environment but should come in as a submodule or other mechanism. The docs for CMAKE/gtest integration (https://github.com/google/googletest/blob/master/googletest/README.md) recommend pulling the dependency from github (I'll submit a patch that uses this method).

pavel-kirienko:
We should abandon GTest and migrate to Catch2 instead, I think

thirtytwobits:
I don't believe catch2 supports mocks. You get gmock for free when you use googletest. Libuavcan would do well to use gmock more. I also submitted a patch to make this less painful.

pavel-kirienko:
I guess now it is a dependency.

PR #133, opened 2018-06-07
#132 proper googletest dependency and fixes for unittests on OSX

thirtytwobits:
Problem:
It's really hard to get Libuavcan tests building on a mac or other platform if googletest isn't installed.

Solution:
add "if linux" guards to problem areas. Also include googletest as recommended in the googletest project's README.

Testing:
Successfully built on OSX sierra using gcc6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ __pycache__
 .pydevproject
 .gdbinit
 
+# vsstudio code
+.vscode
+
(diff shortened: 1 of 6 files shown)


pavel-kirienko:
Thanks. Would you be up to remove GTest from the Travis CI script to see if it still passes after that?

thirtytwobits:
Seems to work.

Issue #100, opened 2017-03-16
Remove dependency on std::auto_ptr in unit tests


pavel-kirienko:
`std::auto_ptr<>` has been removed from newer versions of  STL.

It should be noted that its replacement to newer means of automatic memory management will break the C++03 unit tests. Perhaps the affected tests should be disabled for the C++03 mode. The resulting lack of coverage should be tolerable, especially considering my intention to drop the support for C++03 at some point in the future.

thirtytwobits:
I'd very much like to drop C++03 support. It adds considerable complexity to the codebase and build time to the unittests. Requiring C++11 shouldn't be controversial at all. This has almost universal support at this point.

PR #138, opened 2018-06-14
Issue #100 remove auto_ptr and c++03 support.

thirtytwobits:
Problem:
auto_ptr is deprecated.

Solution:
use unique_ptr which is only available in c++11 and newer.
Also fix how unit tests are run as part of the build to keep them from being
deleted when they fail.

Testing:
Unit tests on linux run.


pavel-kirienko:
Good to merge?

thirtytwobits:
Let's wait until #142 is merged then I'll get this one green.

thirtytwobits:
This is ready now.

pavel-kirienko:
Thanks, Scott!

PR #140, opened 2018-06-15
Feature/issue135 (CAN FD)

thirtytwobits:
Initial swag at CAN-FD support in libuavcan (not including updates to the drivers). This implements a transport CRC scheme that might not be the one we want to go with. This scheme keeps the transport CRC at the beginning of the data transfer and includes buffer bytes (0x55) as part of the CRC.


thirtytwobits:
The tests aren't getting linked correctly using the method I tried to move to. Investigating.

thirtytwobits:
New theory: the Travis hosts are running out of stack memory for some of the tests using FD frames. Let's see if this try fixes it...

pavel-kirienko:
Scott, is this not a good opportunity to separate drivers into several dedicated repositories? Updating all drivers (SocketCAN and STM32) to compatibility with CAN FD seems like a lot of work.

* libuavcan_lpc11c24
* libuavcan_linux
* libuavcan_stm32
* libuavcan_posix_addons (naming ideas welcome)

thirtytwobits:
agreed. I'd also like to build an Atmel/Microchip C21 driver so we can get on-target coverage of CAN-FD.

thirtytwobits:
Should we move the FD work to a branch then? It's obviously not ready for master as the specification doesn't yet support the changes in the PR.

pavel-kirienko:
I've just created a branch [`uavcan-v1.0`](https://github.com/UAVCAN/libuavcan/tree/uavcan-v1.0), let's stage our changes there. We're going to merge it into master once v1.0 is finished and fully supported.

libuavcan/include/uavcan/build_config.hpp
```diff
@@ -220,11 +221,20 @@ namespace uavcan
 static const unsigned MemPoolBlockSize = UAVCAN_MEM_POOL_BLOCK_SIZE;
 #elif defined(__BIGGEST_ALIGNMENT__) && (__BIGGEST_ALIGNMENT__ <= 8)
 /// Convenient default for GCC-like compilers - if alignment allows, pool block size can be safely reduced.
+#ifdef UAVCAN_USE_FD
```
pavel-kirienko: Should be `#if defined(UAVCAN_USE_FD) and UAVCAN_USE_FD` (MISRA 16-0-7). This also applies to all other usages of this and other preprocessor macros for conditional compilation.

libuavcan/include/uavcan/build_config.hpp
```diff
@@ -220,11 +221,20 @@ namespace uavcan
...
+static const unsigned MemPoolBlockSize = 64;
```
pavel-kirienko: 64 bytes are not sufficient for CAN FD - we also need to take into account the CAN ID field and other overheads.

My quick estimation shows that we need (8 bytes pointer + 64 bytes payload + 4 bytes CAN ID + 1 byte DLC + 3 bytes padding + 8 bytes TX deadline + 1 byte QoS + 1 byte IO flags + 6 bytes padding) = 96 bytes large blocks, regardless of the platform. Huh, @davids5 is going to love this.

libuavcan/include/uavcan/build_config.hpp
```diff
@@ -243,8 +253,7 @@ struct UAVCAN_EXPORT IsDynamicallyAllocatable
 {
     static void check()
     {
-        char dummy[(sizeof(T) <= MemPoolBlockSize) ? 1 : -1] = { '0' };
-        (void)dummy;
+        static_assert((sizeof(T) <= MemPoolBlockSize), "Type cannot be allocated dynamically.");
```
pavel-kirienko: So we are officially leaving C++03 behind?

libuavcan/include/uavcan/driver/can.hpp
```diff
@@ -18,6 +19,32 @@ namespace uavcan
  */
 enum { MaxCanIfaces = 3 };
 
+/**
+ * Valid message data lengths.
+ *
+ * CAN DLCs are only 4-bits long so FD had to use 9-15 to encode
+ * various lengths up to 64 bits.
+ */
+enum class MessageDLC : uint_fast8_t {
```
pavel-kirienko: Better type name: `CANFrameDLC` (otherwise there'd be a conflict of terms: in UAVCAN, "message" is a special case of data transfer alongside service request and service response).

You are using scoped enums here, which is great (if we really decided to leave C++03 behind, that is). In this case the prefix `MessageDLC_` becomes redundant: `MessageDLC::MessageDLC_0` - no way.

Actually, come to think of it, do we really need this enum? You have defined two mappings elsewhere which map between data size and DLC and back. Can we not just use these? Or, better yet, there should be constexpr functions for that, for extra compile-time and run-time safety:

```c++
constexpr std::uint8_t lengthToDLC(std::uint8_t length)
{
    UAVCAN_ASSERT(length <= 64);   // Legal since C++14
    return ...; // you get the point
}
```

Another reason this is better than an enum is because enums are not supposed to be casted to integral values. This disagrees with the spirit of MISRA. I understand that there are a lot of placed in libuavcan where I did just that, but that was because I had no other option at the time of writing the original codebase - I had to keep it compatible with C++03 (ugh).

libuavcan/src/transport/uc_transfer_sender.cpp
```diff
@@ -134,6 +140,17 @@ int TransferSender::send(const uint8_t* payload, unsigned payload_len, Monotonic
...
             if (offset >= int(payload_len))
             {
+                for(unsigned i = 0; i < padding_bytes; ++i)
+                {
+                    if(0 == frame.pushBackPayload(BytePaddingPattern))
+                    {
+                        UAVCAN_TRACE("TransferSender", "Frame payload padding failure");
+                        registerError();
+                        // The calculations above should make it impossible
+                        // to get here.
```
pavel-kirienko: These kinds of errors should be accompanied with `UAVCAN_ASSERT(false);` in order to amplify errors in a development environment.

pavel-kirienko:
A few more things to think about. If we were to commit to the migration to C++11 (or even C++14 while we are at it since it has a lot of important fixes and improvements), what are we going to do with all of the outdated and unsafe constructs in the library that were designed to be compatible with C++03? For example, stuff like `enum { CompileTimeConstant = 42 };`, lack of constexpr declarators where appropriate, hand-crafted compile-time logic like `StaticAssert<>` and `EnableIf<>`. Do we just keep it, or do we sweep through the entire codebase replacing those with the alternatives provided by the standard library? What about platforms that *kinda support* C++11 but at the same provide horribly broken standard libraries that belong in a trash can (hello NuttX)? This makes me realize that migration to C++11 is a lot more work than I initially thought, and maybe this is not a good opportunity for that, after all. Maybe it'd be wiser to implement CAN FD support within the current framework supporting C++03, and then tackle its deprecation as a separate commitment?

Another important thing: I know that coding style adherence is always annoying, but we can't do anything about that if we want the codebase to stay clean and readable. Please follow the Zubax Coding Conventions: https://kb.zubax.com/x/84Ah. I decided to desist from listing all of the deviations in this PR, they shouldn't be hard to spot. You could also try this thing, YMMV: https://kb.zubax.com/x/94Ah

When referring to entities from the standard library, always use the `std::` prefix, e.g. `std::size_t`. This is not needed for the standard integral types because libuavcan has to re-export them from within its own namespace in order to support C++03.

pavel-kirienko:
One more thing: I think we should provide some way of switching from CAN FD back to CAN 2.0 at run time when the library is compiled for CAN FD. That will enable compatibility with older systems co-existing on the same bus with CAN FD systems.

libuavcan/include/uavcan/build_config.hpp
thirtytwobits: Yep. Will do. I'll take a look at the Zubax conventions but it'd really help if you had a clang-format file?

libuavcan/include/uavcan/build_config.hpp
thirtytwobits: I kept meaning to ask you about how you calculate this. Can I put your math here into a comment? It'd be great to understand where the initial values came from.

Also: when you say "8 bytes pointer" do you mean two 32-bit pointers or a single 64-bit?

Also, Also: The unit tests ran with this value even under valgrind. How can we write a test that validates this value on a given platform?

libuavcan/include/uavcan/build_config.hpp
thirtytwobits: Yes.

libuavcan/include/uavcan/driver/can.hpp
thirtytwobits: Well, it's a lot safer to cast to integral values in C++11 given that you can specify and safely query the underlying type. That said, I'll give your comments some consideration and see what I can come up with.

thirtytwobits:
## C++11

I think C++11 is a safe choice this year. There are some stragglers getting to C++14 in some of the more exotic RTOS compilers but everyone seems to have been at c++11 for at least a year or two.

Let's iterate on doing a thorough C++11 conversion since it will changes a lot of code. We should pick one construct at a time and change it wholesale. Once I am back from vacation I'll open a series of issues to work through this.

## FD/2.0 switch

Yeah, I thought hard about whether or not allowing a runtime fallback to 2.0 made sense. 

First off, compile-time support for FD is an important feature given the extra memory requirements of the protocol. Either way, we definitely want to support compiling for 2.0 only for platforms with less RAM.

As for also allowing switching at runtime when you do have FD support: yes, I see how this is a useful feature for a "batteries included" implementation like libuavcan but it does add more complexity. It could take some time to cover all the edge cases both in unit tests and with a test rig to validate this on a real bus. Shall I open an Issue to track this feature?

libuavcan/include/uavcan/build_config.hpp
pavel-kirienko: I know but we don't have it. As I just wrote elsewhere, we have a config for Uncrustify, but that tool tends to break formatting occasionally, so YMMV.

libuavcan/include/uavcan/build_config.hpp
pavel-kirienko: > Can I put your math here into a comment?

Sure.

> It'd be great to understand where the initial values came from.

The largest dynamically allocated object type is `uavcan::CanTxQueue::Entry`, so I just pulled the numbers from there.

> Also: when you say "8 bytes pointer" do you mean two 32-bit pointers or a single 64-bit?

No, I mean a single 64-bit pointer, this is the worst case. On a 32-bit system, it would be 32-bit, but the alignment reasons require us to keep the block size aligned at 8 bytes at least (maybe more depending on the platform; 96-byte large blocks can be aligned at 32 bytes, which should suit every sensible use case).

> The unit tests ran with this value even under valgrind. How can we write a test that validates this value on a given platform?

*What no that can't be.* The allocator is designed to fail if you're trying to allocate more bytes than the size of the memory block: https://github.com/UAVCAN/libuavcan/blob/c044630390370d800a22e7746cd949ddd3c59bb1/libuavcan/include/uavcan/dynamic_memory.hpp#L162-L165

pavel-kirienko:
> Let's iterate on doing a thorough C++11 conversion since it will changes a lot of code. We should pick one construct at a time and change it wholesale. Once I am back from vacation I'll open a series of issues to work through this. If we do it this way we can find the specific problem constructs and leave the old implementations in place.

Excellent plan, but what about this specific PR? Should we migrate it back to C++03 in order to keep changing only one thing at a time?

> we definitely want to support compiling for 2.0 only for platforms with less RAM.

Yes.

> It could take some time to cover all the edge cases both in unit tests and with a test rig to validate this on a real bus. Shall I open an Issue to track this feature?

Yes, it can be implemented later perhaps.

pavel-kirienko:
> As for also allowing switching at runtime when you do have FD support: yes, I see how this is a useful feature for a "batteries included" implementation like libuavcan but it does add more complexity.

Come to think of it, this feature is not optional but absolutely necessary. Otherwise, we won't be able to migrate PX4/ArduPilot easily when Pixhawk 5 is released (which will be CAN FD enabled), because there is fielded hardware out there which does not support CAN FD, yet it needs to be compatible with CAN FD enabled autopilots.

libuavcan/include/uavcan/build_config.hpp
thirtytwobits: Remember you merged feature/issue100 already. So the switch has been thrown. It's just a matter of cleaning up now and fully embracing the new language. As I stated before: we should not add more c++11 into this PR and instead add it in one feature at a time when I have a nuttx verification environment setup. So far we are using, and should continue to use, static_assert and unique_ptr (i.e. do not use auto_ptr any more).

To that end: what should I do to be able to verify nuttx compatibility with these changes?

libuavcan/include/uavcan/build_config.hpp
thirtytwobits: I'm a bit confused now that I'm looking at this change. Everything else in this file is using #ifdef. Does this mean you just have a lot of MISRA cleanup or are you ignoring this rule in your coverity scans?

libuavcan/include/uavcan/driver/can.hpp
thirtytwobits: I only find 4-5-2 in MISRA c++ that may have something to say about enumeration values but this prohibits using them with operators other than `[]` which isn't what we are doing with the CanFrameDLC (renamed) type. Unless I'm missing some other rule I don't think my use of the enumeration's values violates any MISRA requirements or advisories.

I do think it's good to use this enumeration since it ensures the compiler catches places where the programmer mixes up the DLC and the data length. Making the two distinct types (i.e. DLC is CanFrameDLC and length is an integer type) give the compiler stricter semantics to work with.

thirtytwobits: So how about:
```
enum class CanFrameDLC : uint_fast8_t {
    code_for_length_0 = 0,       /**< Data Length Code: 0 bytes */
    code_for_length_1 = 1,       /**< Data Length Code: 1 bytes */
   ...
    code_for_length_12 = 9,   /**< Data Length Code: 12 bytes */
    ...
    code_for_length_64 = 15,  /**< Data Length Code: 64 bytes */
    invalid_code = 16 /**< Not a valid DLC. */
};
```

libuavcan/src/transport/uc_transfer_sender.cpp
thirtytwobits: Replacing `UAVCAN_TRACE` with `UAVCAN_ASSERT(false)`

thirtytwobits:
FYI: https://kb.zubax.com/pages/viewpage.action?pageId=2195699#ZubaxC++codingconventions-Preprocessor has a spelling mistake:

`%s/include quards/include guards/`

libuavcan/CMakeLists.txt
```diff
@@ -119,27 +129,36 @@ if (DEBUG_BUILD)
...
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wnon-virtual-dtor")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Woverloaded-virtual -Wsign-promo -Wold-style-cast")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-error=deprecated-declarations")
         set(optim_flags "-O3 -DNDEBUG -g0")
     else ()
         message(STATUS "Compiler ID: ${CMAKE_CXX_COMPILER_ID}")
         message(FATAL_ERROR "This compiler cannot be used to build tests; use release build instead.")
     endif ()
-
+
```
pavel-kirienko: Oh no, I spy a misconfigured text editor.

libuavcan/include/uavcan/build_config.hpp
pavel-kirienko: > what should I do to be able to verify nuttx compatibility with these changes?

Let me invoke @davids5. David, how difficult would it be to set up a testing environment with NuttX on some kind of commonly available STM32H7 demo board like NUCLEO-H743ZI? @LorenzMeier can we get some assistance with that, please?

libuavcan/include/uavcan/build_config.hpp
pavel-kirienko: This applies only to boolean values because a symbol can be explicitly defined as 0 in order to disable some feature; with `#ifdef` this will lead to errors. E.g.:

```c++
#define USE_FEATURE 0
// ...
#ifdef USE_FEATURE
// bug
#endif
```

This also does not apply if we are defining a default value:

```c++
#ifdef USE_FEATURE
# define USE_FEATURE 0
#endif
```

I skimmed though that file again and couldn't find any violations manually.

libuavcan/include/uavcan/driver/can.hpp
pavel-kirienko: Okay. How about making it camel case though?

thirtytwobits: sure but which of these do you want:
1. `codeForLength_n`
3. `codeForLengthN`

pavel-kirienko: `CodeForLengthX`? Or is it PascalCase. I think it's PascalCase. Sorry for the confusion above.

thirtytwobits:
I don't believe the NUCLEO-H743ZI is out yet. Mouser and Digikey still show 0 for their stock. It's rather hard to find a Cortex M with ISO compliant CAN-FD IP right now. The only part I've confirmed that has this is the revision B of the Microchip V71. I believe their C21 and E70 parts are still using the Bosch IP. 

The NXP devkit looks interesting though -> https://www.nxp.com/support/developer-resources/hardware-development-tools/lpcxpresso-boards/lpcxpresso54618-can-fd-kit:OM13094 . The LPC546xx (Cortex M4 core) datasheet claims ISO 11898-1 compliance.

pavel-kirienko:
> I don't believe the NUCLEO-H743ZI is out yet.

Could be - mine still hasn't shipped from Mouser.

> It's rather hard to find a Cortex M with ISO compliant CAN-FD IP right now.

STM32H7 supports ISO 11898 (and it is also compatible with Bosch optionally, there's a software switch).

I am inclined to stay committed to STM32H7 as the main testing target partly because the next Pixhawk is going to be based on that (it is going to support CAN FD they said).

libuavcan/include/uavcan/build_config.hpp
dagar: Plain NuttX or PX4/NuttX?

pavel-kirienko: Plain NuttX preferred, if possible.

kjetilkjeka:
I believe the NXP S32K evaluation board (S32K144EVB) is iso CAN-FD compilant. It's in stock at mouser.

thirtytwobits:
I was able to find a NUCLEO-H743ZI so I'm fine with choosing this as the reference board for FD although I'm partial to the Microchip C21 since it's an M0+ and would keep us more honest. 

The biggest thing blocking progress on this PR is that I can't get the test build for the lpc11c24 to work even using the tip of mainline. When I step through the code I end up in a HardFault_handler right after wwdt_11xx.h:122 (reset the watchdog timer). I'll start diving into the datasheet to understand this chip and figure out where I'm going wrong but at this point I'm not able to verify that my FD change does no harm to a 2.0 build.

pavel-kirienko:
Is this ready for the final review?

thirtytwobits:
No. We need to verify Nuttx compatibility. I'll start working through the px4 developer guide to understand their build system.

thirtytwobits:
Crap. Found a bug that the unittests have somehow missed. We need to quantize the use of MaxBitLen for sizing transfer buffers to account for the padding bytes.

thirtytwobits:
I'm going to be offline until next week. I parked a unit test that reproduces the bug. I'll fix it next week:

https://github.com/UAVCAN/libuavcan/pull/140/commits/88ed4892f4191a8dd04c209dbd35688f6f18036e#diff-78d0f81058d9a7b16d2c06d5bc5ba3faR78

thirtytwobits:
Okay, this should fix the remaining issues (I hope). We still need to test this on Nuttx.

pavel-kirienko:
Would you like to have this merged now, or should we wait until NuttX is tested?

Actually, I think you're better off deciding stuff yourself at this point. I believe you already have all of the necessary permissions to merge things. I have no objections, so merge at will.

thirtytwobits:
I'll probably merge this tonight. I'd wait for nuttx testing but we're on a branch that is not going to be stable for awhile so we might as well embrace some chaos given that this change is getting a bit long in the tooth.

davids5:
@thirtytwobits and @pavel-kirienko - how much of a churn to the hardware drivers does this cause?

scottdarch:
Some. This does introduce changes to the driver layer to support CAN-FD but this is on a branch. We have more work to do in the standard before we will be ready to release this.

That said, @davids5 reminds me that I haven't updated the driver repositories since they didn't exist when I started this patch. I'll do that before merging and I'll put those repos on branches as well.

thirtytwobits:
Yeah, I'm going to merge this into the 1.0 branch now. I need this in so I can finish the driver builds.

PR #143, opened 2018-06-29
Platform drivers moved into dedicated repos

pavel-kirienko:
New home for drivers:

- https://github.com/UAVCAN/libuavcan_linux
- https://github.com/UAVCAN/libuavcan_lpc11c24
- https://github.com/UAVCAN/libuavcan_stm32

I decided to keep the POSIX helper classes in this repo for a want of better ideas.

I strongly suggest to pull this in before continuing with other v1.0 changes.

@thirtytwobits please have a look.


kjetilkjeka:
What is the strategy of keeping the platform driver compatible with changes to the main repo?
For uavcan.rs we use semver together with Cargo. Meaning that cargo will not allow usage for a driver together with a version of uavcan.rs that satisfies the semver properties. 

Do a user have need to both have libuavcan and the platform specific repo as a git submodule?
In uavcan.rs the driver repo will reexport the main library. Meaning that all uavcan stuff is available in the driver namespace as well.

Dependency management in c++ is not as easy as in Rust, so I'm not sure if there is a clear answer to this question. But would it be an idea to have the drivers include libuavcan as a git submodule?

pavel-kirienko:
The dependency management strategy is to let the user manage it manually. The platform driver API is supposed to be **extremely stable**; we are breaking it here now only to support CAN FD, but even considering that, the API shall not change for old hardware that is not CAN FD enabled.

Including libuavcan with drivers is not a good idea as it will impose a lot of additional maintenance burden - whenever the library is updated, the drivers will have to be updated too.

pavel-kirienko:
@kjetilkjeka @thirtytwobits @antoinealb awaiting for your confirmation before merging. After this PR is merged, https://github.com/UAVCAN/libuavcan/pull/140 will require a bit of updating.

PR #151, opened 2018-07-21
Support for nxp20k processors (e.g. Teensy)

fabolhak:
Hello,
thank you for this great library. We think it fits our needs very good and therefore we ported it to the Teensy 3.2 processor which we use. 

The implementation is being tested for the MK20DX256 (Teensy 3.2 and 3.1). However, it may also work with the MK64FX512 (Teensy 3.5) or MK66FX1M0 (Teensy 3.6).

The driver implementation builds upon an existing FlexCAN library:
 https://github.com/collin80/FlexCAN_Library


pavel-kirienko:
Hi Fabian,

Thank you for the contribution. I think it would be fair to mention that the clock implementation is still missing.

We are now migrating to a different arrangement where each platform driver resides in a separate repository:

- https://github.com/UAVCAN/libuavcan_stm32
- https://github.com/UAVCAN/libuavcan_linux
- https://github.com/UAVCAN/libuavcan_lpc11c24

So there is no need to merge your driver (or any other driver) into the main libuavcan repository. If you would like to move your driver into the UAVCAN organization on Github, I suggest fixing the following:

- State explicitly the licensing terms of the FlexCAN library
- Update the code to be compatible with the Zubax Coding Conventions. Avoid pollution of the global scope with macro definitions. Things like `using namespace uavcan;` in the global namespace in header files are absolutely unacceptable. Mind the indentation.
- Provide some docs.

Alternatively, it should be perfectly acceptable to just leave the driver hosted wherever you prefer, since as I said there is now no need to keep the code at the same place. I can link it from uavcan.org as well if you'd like.

Regardless of the way you decided to go, let's move this conversation to the mailing list.

fabolhak:
Hello Pavel,
thank you for the fast reply and detailed response. Yes, licensing is always an issue. Unfortunately, the FlexCan library doesn't state a License which makes it difficult. However, we are very fine with hosting the teensy driver in our forked repository. 

The intent of this PR was more to let you (and anyone who searches for it) know that an early version of the teensy driver exits. I know that it is by far not perfect. However, it works for us and maybe is also helpful for someone else :).

PR #155, opened 2018-07-30
replaced the depricated ifconfig command by the ip command

hjoertel:


libuavcan_drivers/linux/scripts/uavcan_add_vcan
```diff
@@ -19,11 +19,14 @@ if [ "$1" == '--help' ] || [ "$1" == '-h' ]; then echo "$HELP"; exit; fi
 
 IFACE="$1"
 
-if [ $(ifconfig -a | grep -c "^$IFACE") -eq "1" ]; then
-    ifconfig $IFACE up
+ip link show $IFACE > /dev/null
+if [ $? == 0 ]; then
+    ip link set up $IFACE
```
pavel-kirienko: Can we not just say `if ! ip link show $IFACE > /dev/null; then` instead of the two lines above?

libuavcan_drivers/linux/scripts/uavcan_add_vcan
```diff
@@ -19,11 +19,14 @@ if [ "$1" == '--help' ] || [ "$1" == '-h' ]; then echo "$HELP"; exit; fi
...
     exit
 fi
 
+echo  "load driver and start $IFACE"
```
pavel-kirienko: Not sure what it means, could you elaborate, please? Why can't we do this automatically? Everything that can be sensibly automated, should be.

hjoertel:
Am Montag, 30. Juli 2018, 16:58:25 CEST schrieb Pavel Kirienko:
> +echo  "load driver and start $IFACE"
> 
> Not sure what it means, could you elaborate, please? Why can't we do this
> automatically? Everything that can be sensibly automated, should be.

This line only says, that the script is now loading the driver modules  and is 
starting the CAN interface
It doesn't not mean the user has to do it. 
If you like, we can simply remove this line. 

Heinz

pavel-kirienko:
Let's remove it because in its current form it's an imperative, quite confusing.

hjoertel:
Am Montag, 30. Juli 2018, 16:58:25 CEST schrieb Pavel Kirienko:
> +ip link show $IFACE > /dev/null
> +if [ $? == 0 ]; then
> +    ip link set up $IFACE
> 
> Can we not just say `if ! ip link show $IFACE > /dev/null; then` instead of
> the two lines above?

for me it is more clear using it with two lines.
- first issue a command to see if the interface is already there
- second, check the result 

Heinz

pavel-kirienko:
Okay. I will merge this once the confusing echo is removed. Thanks!

hjoertel:
Am Montag, 30. Juli 2018, 19:54:18 CEST schrieb Pavel Kirienko:
> Okay. I will merge this once the confusing echo is removed. Thanks!
OK
do I need to do another pull request?

Heinz

pavel-kirienko:
Thanks! This PR has been updated automatically after you've pushed the second commit. Merging.

PR #158, opened 2018-08-06
Use C++11 std::enable_if instead of our own

### LINKED ISSUE #149: Replace all use of uavcan::EnableIf with std::enable_if
use the standard library's implementation.

antoinealb:
We recently decided to remove support for C++03, which means we can now
use standard library constructs.

Fixes #149  (Replace all use of uavcan::EnableIf with std::enable_if).

### DIFF
--- a/libuavcan/include/uavcan/marshal/array.hpp
+++ b/libuavcan/include/uavcan/marshal/array.hpp
@@ -9,6 +9,7 @@
 #include <cstdio>
 #include <cstring>
 #include <cmath>
+#include <type_traits>
 #include <uavcan/error.hpp>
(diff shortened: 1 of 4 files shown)


pavel-kirienko:
This change might be incompatible with NuttX. We should wait until the NuttX testing environment is available. ✋

thirtytwobits:
While I am admittedly slow to contribute to this project this is my top priority when I do work on it. I too am blocked on more robust changes until I can ensure we haven't already and will not in the future break NuttX compatibility.

pavel-kirienko:
@thirtytwobits @antoinealb Seeing as this codebase will be thrown out anyway, do we still need this PR?

Issue #161, opened 2018-08-28
off-by-one error in CAN-FD changes


thirtytwobits:
Found a logic error in the can.hpp length_to_dlc and dlc_to_length methods where the index is allowed to reference memory after the static LUTs. Fix is forthcoming.

PR #162, opened 2018-08-28
off-by-one error in CAN-FD changes #161

thirtytwobits:


pavel-kirienko:
I would like to point out that the code is not conforming to the coding conventions. The braces are misplaced and the method is named incorrectly.

thirtytwobits:
Too right. Sorry. I'll fix. I like your style better than the one I have to use at work anyway ;-)

Issue #159, opened 2018-08-07
IO latency is bad


pavel-kirienko:
Evidence from @bugobliterator: https://github.com/ArduPilot/ardupilot/pull/8987#issuecomment-409235227 (bad buffering architecture leads to slow field serialization and deserialization)

From @mhkabir:  

> Also after some instrumentation (not related to Myxas directly), I found :
>
> `uavcan control latency: 1268 events, 377221us elapsed, 297.49us avg, min 165us max 2440us 322.559us rms` (Myxa PX4 copter)
>
> `io control latency: 47776 events, 8705549us elapsed, 182.22us avg, min 160us max 291us 49.261us rms` (Normal PX4 copter)
>
> The worst case 2ms latency is b a d.

This is what I had to say on the subject (copy from the linked thread above):

> The linked list based buffer requires sequential iteration of the buffer segments in order to reach the desirable offset, which makes it slow. Additionally, the code that you identified as the culprit makes a lot of auxiliary operations per every byte read in order to facilitate the segment traversal logic at the outer layer. One could get rid of the list traversal problem by using some sort of a stateful reader object which would keep references to the current chunk. Is anybody willing to give it a try?

The culprit can be easily found in the sources by grepping for the key phrase `This shall be optimized`. As it stands, I wrote it back in 2014.

![image](https://user-images.githubusercontent.com/3298404/43795071-e9d6eb0a-9a88-11e8-9f89-650fab6ea13e.png)

pavel-kirienko:
Relevant observations: https://github.com/PX4/Firmware/issues/10171#issuecomment-411669947

thirtytwobits:
I have a fix in the works for this. Stay tuned.

PR #163, opened 2018-09-12
Issue #159 improving IO latency.

thirtytwobits:
We were seeing similar problems when using the file transfer function. This change
improved performance drastically.


libuavcan/src/transport/uc_transfer_buffer.cpp
```diff
@@ -96,18 +96,43 @@ void TransferBufferManagerEntry::Block::destroy(Block*& obj, IPoolAllocator& all
...
-        if (total_offset >= target_offset)
+        const unsigned bytes_available_at_offset = end_of_block - target_offset;
+        bytes_to_copy = bytes_available_at_offset;
+        if (bytes_to_copy > left_to_read)
         {
-            *outptr++ = data[i];
-            left_to_read--;
+            bytes_to_copy = left_to_read;
         }
+        uint8_t *src = data + (target_offset - start_of_block);
```
pavel-kirienko: `uint8_t *src` :scream_cat:

libuavcan/src/transport/uc_transfer_buffer.cpp
```diff
@@ -96,18 +96,43 @@ void TransferBufferManagerEntry::Block::destroy(Block*& obj, IPoolAllocator& all
...
+        ::memcpy(outptr, src, bytes_to_copy);
```
pavel-kirienko: Not to be overly paranoid, but should we not cast the result to `void` to keep compliance with MISRA?

I am aware that the codebase has other non-MISRA-compatible features, such as the use of plain native types like `unsigned` instead of deterministic `uint32_t`. Hopefully that will be resolved one day, too.

libuavcan/src/transport/uc_transfer_buffer.cpp
```diff
@@ -96,18 +96,43 @@ void TransferBufferManagerEntry::Block::destroy(Block*& obj, IPoolAllocator& all
...
+    // Case 2 -- start of wanted area was before this block
+    else
+    {
+        UAVCAN_ASSERT(target_offset < start_of_block);
+        bytes_to_copy = Block::Size;
+        if (bytes_to_copy > left_to_read)
+        {
+            bytes_to_copy = left_to_read;
+        }
+        ::memcpy(outptr, data, bytes_to_copy);
```
pavel-kirienko: Same here

libuavcan/src/transport/uc_transfer_buffer.cpp
```diff
@@ -165,17 +190,40 @@ int TransferBufferManagerEntry::read(unsigned offset, uint8_t* data, unsigned le
     }
     UAVCAN_ASSERT((offset + len) <= max_write_pos_);
 
-    // This shall be optimized.
     unsigned total_offset = 0;
     unsigned left_to_read = len;
     uint8_t* outptr = data;
-    Block* p = blocks_.get();
+    Block* p;
```
pavel-kirienko: `Block* p = nullptr;`

libuavcan/src/transport/uc_transfer_buffer.cpp
```diff
@@ -165,17 +190,40 @@ int TransferBufferManagerEntry::read(unsigned offset, uint8_t* data, unsigned le
...
+
+    // Check if this data starts in the previously cached block,
+    // if so, use it.
+    if (previous_block_ && (offset >= previous_block_start_offset_ && offset < (previous_block_start_offset_ + Block::Size)))
```
pavel-kirienko: The braces on the right side are placed awkwardly. Either the right side braces are not required or it's a bug.

libuavcan/src/transport/uc_transfer_buffer.cpp
thirtytwobits: On this point, I'm not yet able to submit coverity builds. Do we have that scan setup to do MISRA when it does work?

pavel-kirienko: IIRC last time I checked MISRA-specific checks were not available on the free plan that we're using. So Coverity is configured to run only general checks.

thirtytwobits: What MISRA rule is this a violation of? I see that 5-0-15 of MISRA C++ allows pointer arithmetic when calculating array indices.

thirtytwobits: ah, done.

libuavcan/src/transport/uc_transfer_buffer.cpp
pavel-kirienko: What about the coding style violations though. I now understand that my previous comment might be overly cryptic, but what I was hinting at is that the asterisk is misplaced. I recall you made a config file for Clang autoformatter, maybe it is time to use it here?

thirtytwobits:
Sorry about the force pushes. You're right. The process at my day-job is different then github so I forget this sometimes. Won't do this in the future.

Issue #164, opened 2018-09-19
Add "get adjustment" method to ISystemClock


thirtytwobits:
When working on improving the SystemClock abstraction in libuavcan_linux we ran into an issue where the core driver depended on an implementation-specific method, getPrivateAdjusment. To allow creation of drivers that only use ISystemClock we would like to add a generic version of this method to the ISystemClock interface.

pavel-kirienko:
I don't like this. `getPrivateAdjustment()` is a bit of a hack, needed to work around the case where the node needs to be time-synced with the network, while not changing the system-wide clock settings. The whole concept of it is very specific to the Linux platform driver. Can you not just add a lower-level interface underneath `ISystemClock`, say, `IPrivateSystemClock`?

thirtytwobits:
Take a look at my PR and see if it makes sense when you read the comments?

pavel-kirienko:
```
     * Note that this value is not necessarily the value set via adjustUtc(UtcDuration) but may be 
     * related for some implementations. For example, where a system clock implements adjustUtc(UtcDuration)
     * by adjusting the actual system time this method will always return 0 since time obtained from this 
     * system no longer needs to be adjusted. The opposite is true where an implementation cannot adjust 
     * the system time and so must always manually apply this adjustment to any value retrived from
     * the system.
```

What I'm reading is that the behavior would still be implementation-defined, but the interface will be unified. Why not express it clearly by adding a dedicated interface?

thirtytwobits:
because it's expressing how the ISystemClock::getUtc() method is determining its value. The two methods are closely related and so part of the same interface (IMHO).

pavel-kirienko:
> because it's expressing how the ISystemClock::getUtc() method is determining its value.

But that's not quite the case. The interface does not put any constraints on the way `getUtc()` operates. I think adding this method is undesirable because it undermines encapsulation, and it can be easily abused to create implementations tightly coupled with a particular implementation of the clock driver, which is undesirable.

thirtytwobits:
Okay. So how about ISystemClockAdjustment then?

pavel-kirienko:
Could you please elaborate a bit on the overall objective?

> To allow creation of drivers that only use ISystemClock we would like to add a generic version of this method to the ISystemClock interface.

As I understand it, you want a SocketCAN driver that doesn't use the Linux-specific implementation of ISystemClock?

thirtytwobits:
[...]

PR #165, opened 2018-09-19
Issue #164 Proposed new API

thirtytwobits:
For discussion. Doesn't compile.


thirtytwobits:
Per comments, moving change down into the driver package.

Issue #74, opened 2015-10-20
POSIX storage backend for dynamic node ID allocation server fails to initialize if the directory doesn't exist


pavel-kirienko:
Discovered with Linux dynamic node ID allocation application: if the storage directory doesn't exist when the application is starting, the backend will create the directory and then fail. On the second attempt it always works fine.

```
Error: libuavcan/libuavcan_drivers/linux/apps/uavcan_dynamic_node_id_server.cpp:495: 0 <= storage_backend.init(persistent_storage_path.c_str())
```

@davids5 FYI.

edwinhayes:
Is this going to be fixed, or is this supposed to be considered normal?

pavel-kirienko:
This is a low priority issue. Somebody should fix it. Please fix it.

PR #169, opened 2018-10-16
Fix path creation when parent doesn't exist

SyrianSpock:
Fixes the issue reported in #74. This was caused by the failure of `mkdir` to create the specified directory because leading directories didn't exist. The patch proposed creates leading directories from top to bottom to ensure the whole tree is setup properly.


pavel-kirienko:
Finally! Thank you, Syrian Spock 🖖

SyrianSpock:
Happy to help 🖖

PR #170, opened 2018-10-21
Merge back from mainline ahead of resuming v1.0 development

thirtytwobits:
Most of these changes were elided since the drivers were removed from the repo in this branch. Making this a full merge regardless so we can see that all changes were considered.


thirtytwobits:
Ack. Nevermind. I need to do this a cherry-picks. Closing.

PR #188, opened 2018-12-03
Adding coveralls support

thirtytwobits:
Not sure if this actually works but the only way to try it out is to merge (because secure keys aren't available to pull-requests)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,9 @@ vagrant ssh
 mkdir build && cd build && cmake .. -DCMAKE_BUILD_TYPE=Debug -DCONTINUOUS_INTEGRATION_BUILD=1
 ```
 
-> Note that -DCONTINUOUS_INTEGRATION_BUILD=1 is required for this build as the realtime unit tests will not work on a virt.
+Note that you may (probably will) have to increase the virtual memory available to the virtual machine created by Vagrant.
+
+Also note that -DCONTINUOUS_INTEGRATION_BUILD=1 is required for this build as the realtime unit tests will not work on a virt.
 
 You can build using commands like:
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -136,21 +137,6 @@ or to run a single test:
...
-cmake ../../libuavcan -G"Eclipse CDT4 - Unix Makefiles" \
-                      -DCMAKE_ECLIPSE_VERSION=4.3 \
-                      -DCMAKE_BUILD_TYPE=Debug \
-                      -DCMAKE_CXX_COMPILER_ARG1=-std=c++11
-```
-
-Path `../../libuavcan` in the command above points at the directory where the top-level `CMakeLists.txt` is located;
-you may need to adjust this per your environment.
-Note that the directory where Eclipse project is generated must not be a descendant of the source directory.
-
```
pavel-kirienko: why remove though

thirtytwobits: Because it's outdated, I've never tested it, and I don't want to support it. I used Eclipse for years but I gave up on it about six months ago. That project is such a hot-mess and vscode is almost at parity for C++ now so if we are going to provide help for an editor it should be that one. Frankly I'd rather just remain IDE friendly but agnostic. No information is better then wrong information.

pavel-kirienko: image

pavel-kirienko: I've been using CLion lately, it's mostly harmless and its embedded indexer can actually parse most of C++ correctly. Not sure about vscode. The question of "why aren't there any decent C++ IDE in 2018" keeps me up at night though.

README.md
```diff
@@ -122,7 +122,8 @@ vagrant ssh
 mkdir build && cd build && cmake .. -DCMAKE_BUILD_TYPE=Debug -DCONTINUOUS_INTEGRATION_BUILD=1
 ```
 
-> Note that -DCONTINUOUS_INTEGRATION_BUILD=1 is required for this build as the realtime unit tests will not work on a virt.
+> Note that you may (probably will) have to increase the virtual memory available 
+> Also note that -DCONTINUOUS_INTEGRATION_BUILD=1 is required for this build as the realtime unit tests will not work on a virt.
```
pavel-kirienko: `>` are used to denote quotes, this seems like a misuse?

PR #193, opened 2019-01-18
Add generic subscriber recieve filtering

mattofak:
Sometimes it's possible to save some work by deciding before the message
is decoded to not decode it. Do this having a callback function that
only takes generic metadata about the message.

The use case this is specifically being added for is when there are
multiple UAVCAN node classes listening to different sources of data and
the application needs to chose which node it is actually going to listen
to.

### DIFF
--- a/libuavcan/include/uavcan/node/generic_subscriber.hpp
+++ b/libuavcan/include/uavcan/node/generic_subscriber.hpp
@@ -16,6 +16,10 @@
 #include <uavcan/marshal/scalar_codec.hpp>
 #include <uavcan/marshal/types.hpp>
 
+#if UAVCAN_CPP_VERSION >= UAVCAN_CPP11
+# include <functional>
+#endif
+
 namespace uavcan
 {
 /**
@@ -129,6 +133,12 @@ class UAVCAN_EXPORT GenericSubscriber : public GenericSubscriberBase
 {
     typedef GenericSubscriber<DataSpec, DataStruct, TransferListenerType> SelfType;
 
+#if UAVCAN_CPP_VERSION >= UAVCAN_CPP11
+    typedef std::function<bool (const IncomingTransfer&)> DecodeFilter;
+#else
+    typedef bool (*)(const IncomingTransfer&) DecodeFilter;
+#endif
+
     // We need to break the inheritance chain here to implement lazy initialization
     class TransferForwarder : public TransferListenerType
     {
@@ -153,6 +163,8 @@ class UAVCAN_EXPORT GenericSubscriber : public GenericSubscriberBase
     };
 
     LazyConstructor<TransferForwarder> forwarder_;
+
+    DecodeFilter decode_filter_ {};
 
     int checkInit();
 
@@ -217,6 +229,16 @@ protected:
     }
 
     TransferListenerType* getTransferListener() { return forwarder_; }
+
+ public:
+    /**
+     * Set filter function determining if a given transfer will be decoded.
+     * By default, or if \p filter is nullptr, all messages will be decoded.
+     */
+    void setDecodeFilter(DecodeFilter filter)
+    {
+      decode_filter_ = filter;
+    }
 };
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
This is interesting, thank you.

> Sometimes it's possible to save some work by deciding before the message is decoded to not decode it.

Please name a few cases where this is possible without looking inside the message. Also, are we sure that the overhead of checking whether the decode handler is installed when it's not is worth its benefits?

(this version of the codebase is no longer actively maintained, we are doing a full rewrite; the effort is coordinated by @thirtytwobits )

mattofak:
In my particular case I have multiple communication buses for redundancy. Each bus contains the same message IDs, but not necessarily the same message contents as the originating processors differ. My application needs to be able to choose what bus to listen to on a per message basis. In this simplistic case, merely knowing what bus a message came in on and it's ID is sufficient.

I could do this in the receive handler after the message has been decoded; but then each handler would have to contain a call to the filter. This was made more difficult by the fact that `ReceivedDataStructure` does not have direct access to the `IncomingTransfer` struct. I considered making an interface for the getter methods so that I could pass that to a generic filter; but in the end this appeared cleaner as my manager class can take references to `GenericSubscriber`s, not have to care about the `DataType_` being received, and not have to trouble the actual receiving code with the voting logic happening behind the scenes.

> Also, are we sure that the overhead of checking whether the decode handler is installed when it's not is worth its benefits?

Given that I may be the first person to require this behavior, possibly not. But the addition of a single branch doesn't seem too onerous in the grand scheme of things. My comment about performance was probably not useful anyways because my motivation for placing this here was more about management of the filters and any efficiency gains from not decoding the message were a nice extra.

mattofak:
Regarding the rewrite; I'm happy to leave this change downstream if you wish not to take it. But it may be useful information in general to know that there are use cases where the communications buses are not completely identical. Not only do I have a need to selectively receive messages based on bus, but I also have a need to selectively send messages to buses. That behavior I currently implement by wrapping the publisher class in one of my own doing.

libuavcan/include/uavcan/node/generic_subscriber.hpp
```diff
@@ -252,14 +274,21 @@ int GenericSubscriber<DataSpec, DataStruct, TransferListenerType>::checkInit()
...
+        return;
+    }
 
     /*
      * Decoding into the temporary storage
      */
+    ReceivedDataStructureSpec rx_struct(&transfer);
     BitStream bitstream(transfer);
     ScalarCodec codec(bitstream);
-
```
pavel-kirienko: Please restore the blank line.

libuavcan/include/uavcan/node/generic_subscriber.hpp
```diff
@@ -217,6 +229,16 @@ class UAVCAN_EXPORT GenericSubscriber : public GenericSubscriberBase
...
     TransferListenerType* getTransferListener() { return forwarder_; }
+
+ public:
+    /**
+     * Set filter function determining if a given transfer will be decoded.
+     * By default, or if \p filter is nullptr, all messages will be decoded.
+     */
+    void setDecodeFilter(DecodeFilter filter)
+    {
+      decode_filter_ = filter;
```
pavel-kirienko: Bad formatting.

libuavcan/include/uavcan/node/generic_subscriber.hpp
```diff
@@ -217,6 +229,16 @@ class UAVCAN_EXPORT GenericSubscriber : public GenericSubscriberBase
     }
 
     TransferListenerType* getTransferListener() { return forwarder_; }
+
+ public:
+    /**
+     * Set filter function determining if a given transfer will be decoded.
+     * By default, or if \p filter is nullptr, all messages will be decoded.
```
pavel-kirienko: It should be clarified here that if the predicate returns false, the transfer will be dropped (ignored).

libuavcan/include/uavcan/node/generic_subscriber.hpp
```diff
@@ -154,6 +164,8 @@ class UAVCAN_EXPORT GenericSubscriber : public GenericSubscriberBase
 
     LazyConstructor<TransferForwarder> forwarder_;
 
+    DecodeFilter decode_filter_ {};
```
pavel-kirienko: Extra space before `{}`.

pavel-kirienko:
Hey @mattofak, are you going to continue with this, or should we close this for now?

PR #196, opened 2019-01-23
Documenting top-level folders and targets.

thirtytwobits:
Adding documentation for how the top-level folder structure should work
for the bluesky branch. Also adding guidance for target support we plan
on adding to the project as part of the initial release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,3 +79,25 @@ vagrant ssh -c "cd /vagrant/build && make libuavcan_test && ./libuavcan/libuavca
 ### Submitting a Coverity Scan build
 
 We'll update this section when we enable Coverity builds for the blue-sky implementation.
+
+### Folder Structure
+
+**/libuavcan/include** - Contains the entire header-only libuavcan library.
+
+**/test/native** - Unit-tests that validate the libuavcan library. These tests compile and execute using the build host's native environment. They also do not require any communication interfaces, virtual or otherwise, from the operating system and have no timing constraints.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,3 +79,21 @@ vagrant ssh -c "cd /vagrant/build && make libuavcan_test && ./libuavcan/libuavca
...
+
+**/libuavcan/include** - contains the entire head-only libuavcan library.
+
+**/test/native** - a set of gtest unit-tests that validate the libuavcan library. It is setup to compile and execute using the build host's native environment. These tests do not require any actual interfaces, virtual or otherwise, from the operating system and have no timing requirements. Their only requirement should be google test's requirements.
+
+**/test/ontarget** - a set of tests designed to be cross-compiled for a known set of reference hardware and run on a set of dedicated test devices. These tests may have strict timing requirements and may require specific busses and other test apparatuses be present. Each on-target set will fully document such conditions.
+
+**/example** - contains a set of example applications providing real, practical, and tested uses of libuavcan for various targets.
+
+#### Supported targets
```
pavel-kirienko: I recall we [had this conversation](https://forum.uavcan.org/t/v1-reference-hardware/254/3?u=pavel.kirienko) before: can we please avoid stating preference for any particular type of hardware, at least outside of the test suite? This might be damaging to the project I'm afraid. If you want to say that the test suite is supposed to run on the S32K146 demoboard, SocketCAN, and PX4, that's fine; but it should not be implied that the library is officially compatible with those targets only.

thirtytwobits: Okay. I did mean this but I'll try to clarify. It is important, _for testing_, that any contributor can test on-target and that any on-target errors can be reproduced.

PR #200, opened 2019-01-30
In which we argue about build flags.

thirtytwobits:
Time to have this argument.
Also: decided to continue with #ifdef include guards unless you want to
have the pragma argument too?

Other decisions in here:
- bionic as a reference build environment for CI (because xenial is EOL in two years)
- cmake 3.5.1 because that's the default in bionic and it's not all that new either.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,77 @@
+#
+# Copyright (C) 2014 Pavel Kirienko <pavel.kirienko@gmail.com>
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+
+cmake_minimum_required(VERSION 3.5.1)
+
+project(libuavcan CXX)
+
+#
+# Framework : Googletest
+#
+# (Taken from googletest/README.md documentation)
+# GTest executables
+# Download and unpack googletest at configure time
+configure_file(CMakeLists.txt.in googletest-download/CMakeLists.txt)
+execute_process(COMMAND ${CMAKE_COMMAND} -G "${CMAKE_GENERATOR}" .
+    RESULT_VARIABLE result
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/googletest-download )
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fprofile-arcs -ftest-coverage")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Og -DNDEBUG -ggdb -fno-strict-aliasing")
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -pedantic -Wfloat-equal -Wconversion -Wabi")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wunused-parameter -Wunused-variable -Wunused-value")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wcast-align -Wmissing-declarations -Wno-missing-field-initializers ")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wdouble-promotion -Wswitch-enum -Wtype-limits -Wno-error=array-bounds")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wzero-as-null-pointer-constant -Wnon-virtual-dtor")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Woverloaded-virtual -Wsign-promo -Wold-style-cast")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-error=deprecated-declarations")
```
pavel-kirienko: Missing `-Wsign-conversion`?

Why `-Wno-error=deprecated-declarations`? This is a clean-slate codebase, not like we have to tolerate some ugly legacy.

PR #206, opened 2019-02-13
Many fixes to CMakeLists and dsdl compiler

thirtytwobits:
Modenizing the build script and making the dsdl compiler more flexible.

Switching to pydsdl for code generation

This is a partial fix. The generator is junk now and needs a complete
rewrite. This change does prove out the makefile and interface to the
generator.

What's important is to agree on, without question, my decisions about submodule, commons, and virtualenv. Dissent will not be tolerated.

I wouldn't spend any time reviewing libuavcan_dsdl_compiler/__init__.py. I hacked it to simply spew .hpp files whether they were valid C++ or not. I have a change to rewrite dsdlc (I'm calling mine dsdl2cpp) in the works so this code is about to be deleted.

Some concepts introduced in libuavcan_dsdlc with this change will carry forward in dsdl2cpp:

1. --dry-run will allow the build system to enumerate outputs obviating the need for a "stamp" file.
2. template code will be located along side the cpp source under libuavcan/. This is because the template must remain in-sync with the source but the generator has some flexibility in how it can change without necessarily breaking the templates.
3. the generator will not try to detect rebuilds (it will always generate). Build systems like make do this for us and are better at it.

I'm also considering a switch to jinja2 for a template language. I have more experience with jinja2 and I find the templates more readable. But this is a debate for my next code review!


libuavcan/include/uavcan/introspection.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "uavcan/uavcan.hpp"
+
+
+#ifndef UAVCAN_INTROSPECTION_TRACE_ENABLE
+# define UAVCAN_INTROSPECTION_TRACE_ENABLE 0
+#endif
+
+/*
+ * TODO: find a way to do this using SFINAE. Minimize C-preprocessor usage.
```
pavel-kirienko: I doubt it's doable without the preprocessor, since SFINAE depends only on the supplied or deduced template arguments, meaning that the invoking code will have to select the appropriate version at the time of instantiation. The only possibility that comes to mind is to provide a macro the value of which alternates between traced and non-traced builds and use it in place of a template argument, but then again, how is it better than what you already have here -- we'd still have to rely on the preprocessor. I may be missing something though. It's been a while since I've done any serious C++-fu myself.

pavel-kirienko:
> 3\. the generator will not try to detect rebuilds (it will always generate). Build systems like make do this for us and are better at it.

I highly recommend implementing lazy rewrite like it was done in the original transpiler: if the newly generated code does not differ from the existing file, it is not rewritten, so that dependent includees are not recompiled. Unless I am misunderstanding your intention, it may simplify the build setup drastically. The transpiler itself is extremely fast, so we save nothing by skipping its runs.

https://github.com/UAVCAN/libuavcan/blob/ee272fd48b614da92d8532527d61ebca55c561d2/libuavcan/dsdl_compiler/libuavcan_dsdl_compiler/__init__.py#L118-L124

> I'm also considering a switch to jinja2 for a template language. I have more experience with jinja2 and I find the templates more readable. But this is a debate for my next code review!

Think about dependencies though. Pyratemp is a one-file MIT-licensed redistributable library, which is just wonderful. Jinja2 may end up being an external dependency, which is **bad** for libuavcan - you'd have to install additional uncommon stuff just to build the library. Unless we're going to add jinja2 as a submodule, of course, but then again, who wants additional submodules.

libuavcan/include/uavcan/build_config.hpp
```diff
@@ -1,16 +1,64 @@
...
+ *
+ * This config automatically sets according to the actual C++ standard used by the compiler.
+ *
+ * libuavcan v2.0 relies on the standard library and requires features introduced in c++11.
+ * While the library does not require standards newer then c++11 it does compile correctly 
+ * in code that compiles using the newer 2014 and 2017 standards.
+ */
+#define UAVCAN_CPP11    2011
+#define UAVCAN_CPP14    2014
+#define UAVCAN_CPP17    2017
```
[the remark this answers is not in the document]
thirtytwobits: Sure.

libuavcan/include/uavcan/introspection.hpp
thirtytwobits: Okay. I was just writing notes to my future self here anyway. Let's see what pans out when I get here.

thirtytwobits:
Replacing with another PR.

Issue #187, opened 2018-11-19
rename "driver" layer to "platform"


thirtytwobits:
Given that the stuff under "driver" in libuavcan is normally implemented to use a driver (i.e. it is not itself a driver) using the term can be particularly confusing. As such we should acknowledge that interfaces like ICanIface may or may not be implemented as a driver but may sit on top of one (Indeed ICanIface actually sits several layers above a CAN driver when used on top of socketCAN).

thirtytwobits:
This is in the v1 branch (i.e. "bluesky")

Issue #176, opened 2018-11-01
UTC is a misnomer


pavel-kirienko:
There are two types of clock in libuavcan:

* Monotonic clock, which only goes forward steadily at the same rate (with possible minor rate deviations caused by the instability of the base oscillator), and never leaps.
* Synchronized clock, which can leap and change rate in order to keep the local time base synchronized with external sources.

Now, the current naming convention is as follows: the monotonic clock is called "monotonic", as one would expect, but the synchronized clock is called "UTC". I guess my mind was in a serous disarray when I came up with that naming; we have to do something about that.

I suggest replacing "UTC" with "Synchronized", and keep "UTC" as a deprecated alias for a while.

antoinealb:
I don't have a strong opinion on this, but as a point of data, POSIX uses "Monotonic" and "Realtime" to describe those two types of clocks: https://linux.die.net/man/3/clock_gettime

kjetilkjeka:
Realtime may be misleading. The rust stdlib uses "SystemTime" for this type of clock, I think that conveys more information and is less confusing than "Syncrhonized"

antoinealb:
I think SystemTime is pretty good indeed.

pavel-kirienko:
> "SystemTime" [...] conveys more information and is less confusing than "Syncrhonized"

I would argue the direct opposite of that. System time is far too generic, could mean anything, including the monotonic time (remember this is not a local OS we're talking about, but a protocol). If we fail to agree on this, how about saying "Wall time"? This term is used in ROS.

SergeyF1987:
Synctonic will be lol)

pavel-kirienko:
After some extra consideration I ended up agreeing with Kjetil that System Time might be an acceptable option. It is used in the same sense in other sources, including standard libraries of C++ or Wolfram Mathematica. It should also be acceptable to extend the notion of "system" from just the local machine running a UAVCAN node to the entirety of the bus.

Unless there are objections, let's choose Monotonic and System.

antoinealb:
> It should also be acceptable to extend the notion of "system" from just the local machine running a UAVCAN node to the entirety of the bus.

Actually that makes me realize, why not call them `Monotonic` and `NetworkTime` (or `BusTime`). That being said, I am OK with System.

thirtytwobits:
Do we want to fix this in v0 as well @pavel-kirienko ?

pavel-kirienko:
I don't think we should do that.

PR #212, opened 2019-05-01
Many fixes to CMakeLists and dsdl compiler

thirtytwobits:
Massive integration here, sorry. These changes are to kick start CI builds and other formal mechanisms which will govern on going development. Included are:

## Goodies

1. **Full docker support** – [uavcan/libuavcan](https://cloud.docker.com/u/uavcan/repository/docker/uavcan/libuavcan) is now a container in our dockerhub repository. I also have a .travis.yaml file that should enable CI builds using this container. Once this PR is merged I'll enable travis builds for the bluesky branch and will probably have a PR or two to get it all working.
1. **CMake scripts to support multi-target builds** – ci.sh will build "native" (i.e. whatever "this" platform is) and arm-none-eabi.
1. <strike>**A working ST32H _build_ of our unit tests** –  The tests don't yet actually work on target but they are compiling and we have very little work left to make this a complete solution. I'm currently working with an NXP FAE to enable this on the S32K146 per our plans to use that as our official embedded test platform. When I do that I'll remove the ST32H files.</strike>
1. **A working S32K146 build of our unit tests** – This build produces binaries that can be loaded and run on a S32K146EVB and will report gtest results on LPUART1.
1. **Arm-none-eabi build of googletest/googlemock** – It can be done. It has been done.
1. **pydsdlgen integration**
1. **doxygen build**
1. **no submodules** – sorta. Switched to using CMake external projects for our external dependencies (dsdl types and googletest). This means that after a clone you only have to `mkdir build; cd build, cmake ..` to get a full set of working makefiles and all our dependent sources.
1. **lcov integration started** – but not completed. I'll make this a priority to get coverage reports and codacity integration allowing us to monitor the quality of the ongoing work.
1. **clang-format build-check** – Builds fail on violations of our .clang-format rules.
1. **saturating add/subtract supporting all stdint types without compiler intrinsics** – This logic should be portable to all architectures that use twos complement to represent signed integers. We may decide to support optional intrinsics where available (e.g. on arm we could use SSAT USSAT for everything up to (u)int32. We'd need to fall back to our software implementation for (u)int64).
1. **parameterized tests for libuavcan** – This change establishes parametrized tests as a set of first-class test artifacts we will export with the library. Users can chose to implement these for their target to verify any code that may be sensitive to different compilers and/or architectures. See libuavcan/include/test/ptest_time.hpp for our first such artifact. The idea is to allow platform verification suites to be developed in our platform specific driver repositories.
1. **Scott thinks he understands cmake now** – but is still unsure why.

## Still TODO

Things I still want to accomplish from a build and test standpoint are:

1. **clang-tidy** – Build-time linting.
1. **on-target test automation** – Finish the "flash, run, report" part of the on-target test story. This may also include switching to buildkite as a CI provider to allow for connected hardware.
1. **doxygen warnings as errors** – Get the public headers cleaned up enough to force us to follow our documentation rules.
1. **coverage reports** – Mentioned above: lcov reporting to codacity.
1. **documentation hosting** – Find a solution for automatically pushing doxygen HTML to a public server.


cmake/compiler_flag_sets/cortex-m4-fpv4-sp-d16-nosys.cmake
```diff
@@ -0,0 +1,97 @@
...
+                "-pedantic"
+                "-Wall"
+                "-Wextra"
+                "-Werror"
+                "-Wfloat-equal"
+                "-Wabi"
+                "-Wunused-parameter"
+                "-Wunused-variable"
+                "-Wunused-value"
+                "-Wno-missing-field-initializers"
```
pavel-kirienko: Why?

cmake/compiler_flag_sets/native.cmake
```diff
@@ -0,0 +1,74 @@
...
+                "-Werror"
+                "-Wfloat-equal"
+                "-Wconversion"
+                "-Wabi"
+                "-Wunused-parameter"
+                "-Wunused-variable"
+                "-Wunused-value"
+                "-Wcast-align"
+                "-Wmissing-declarations"
+                "-Wno-missing-field-initializers"
```
pavel-kirienko: Why?

cmake/compiler_flag_sets/native.cmake
```diff
@@ -0,0 +1,74 @@
...
+                "-Wswitch-enum"
+                "-Wtype-limits"
+                "-Wno-error=array-bounds"
+)
+
+set(CXX_FLAG_SET ${C_FLAG_SET})
+set(ASM_FLAG_SET ${C_FLAG_SET})
+
+#
+# General C-only flags
```
pavel-kirienko: Do we actually need those?

cmake/modules/clang-format-check.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python3
```
pavel-kirienko: FWIW there is a project I am involved in where the style is enforced as follows:

```bash
clang-format -i -fallback-style=none -style=file $all_source_files || exit 1
modified="$(git status --porcelain --untracked-files=no)"
if [ -n "$modified" ]; then echo "Run enforce_style.sh to reformat the code."; exit 1; fi
```

It doesn't fit the description "no source code is modified", but as we're doing that in the CI environment we don't really care. Just sharing here.

libuavcan/include/test/ptest_time.hpp
```diff
@@ -0,0 +1,88 @@
...
+ */
+/** @file */
+#ifndef UAVCAN_TEST_TIME_HPP_INCLUDED
+#define UAVCAN_TEST_TIME_HPP_INCLUDED
+
+#include "uavcan/uavcan.hpp"
+#include "uavcan/time.hpp"
+#include "gtest/gtest.h"
+
+namespace uavcan
```
pavel-kirienko: Oh, this reminds me. Suppose you're building an application which has a UAVCAN node in it (I mean, can you imagine?). You want the node to be stowed away into its very own namespace. How would you call it?

Say, **`uavcan`**? Oops, this one's already taken.

Should we not rename the library's namespace into `libuavcan`? Or would that be too long?

libuavcan/include/test/ptest_time.hpp
```diff
@@ -0,0 +1,88 @@
...
+    TypeParam instance = TypeParam::fromUSec(std::numeric_limits<int64_t>::min()) - TypeParam::fromUSec(1);
+    ASSERT_EQ(std::numeric_limits<int64_t>::min(), instance.toUSec());
+}
+
+// clang-format off
+REGISTER_TYPED_TEST_SUITE_P(DurationTest,
+                            DefaultValue,
+                            Concept_fromUSec,
+                            SaturatedAdd,
+                            SaturatedSubtract);
```
pavel-kirienko: Turning off clang-format is dangerous and awkward. Have you considered doing this instead:

```c++
REGISTER_TYPED_TEST_SUITE_P(DurationTest,  //
                            DefaultValue,
                            Concept_fromUSec,
                            SaturatedAdd,
                            SaturatedSubtract);
```

libuavcan/include/uavcan/bus/can.hpp
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * The maximum size of a data frame for this bus.
+ */
+constexpr static size_t max_frame_size_bytes = 64;
+
+/** The size of the tail byte, in bytes. */
+constexpr static size_t tail_byte_size_bytes = 1;
+
+/** The number of bytes in the transfer CRC. */
+constexpr static size_t transfer_crc_size_bytes = 2;
```
pavel-kirienko: What is the point of specifying the CRC size and tail bytes on the per-transport basis? These parameters are identical for both CAN transports, and other transports are far more different than can be described using this mechanism.

libuavcan/include/uavcan/bus/can.hpp
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @namespace TypeFd
+ * Properties of an ISO compliant CAN-FD bus.
+ */
+namespace TypeFd
+{
+/**
+ * The maximum size of a data frame for this bus.
+ */
+constexpr static size_t max_frame_size_bytes = 64;
```
pavel-kirienko: Bad naming, should be PascalCase.

Should those not be contained inside a struct for easier grouping when used in templates?

libuavcan/include/uavcan/bus/can.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+}  // end namespace Type2_0
+
+/**
+ * Bit pattern to fill padding bytes with. The UAVCAN specification does not mandate this value
+ * and the actual value of padding bytes must be ignored when receiving messages.
+ * When transmitting use this pattern to minimize the number of stuff bits added by the
+ * CAN hardware.
+ */
+constexpr uint8_t bytePaddingPattern = 0x55;
```
pavel-kirienko: Bad naming again. I should avoid reporting trivial issues like this one from here on.

libuavcan/include/uavcan/porting/clock.hpp
```diff
@@ -0,0 +1,66 @@
...
+     * It doesn't have to be UTC, the name is a bit misleading - actual time base doesn't matter.
+     *
+     * This clock can be synchronized with other nodes on the bus, hence it can jump and/or change
+     * rate occasionally.
+     * This clock is optional; if it is not supported, return zero. Also return zero if the UTC time
+     * is not available yet (e.g. the device has just started up with no battery clock).
+     *
+     * For POSIX refer to clock_gettime(), gettimeofday().
+     */
+    virtual UtcTime getUtc() const = 0;
```
pavel-kirienko: We should probably look into resolving https://github.com/UAVCAN/libuavcan/issues/176 after this PR is merged?

libuavcan/include/uavcan/time.hpp
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * @tparam DURATIONTYPE  The datatype returned when retrieving durations from
+ *                       realizations of this base class. This type must be
+ *                       exactly 8 bytes in size and must support certain concepts
+ *                       implied by DurationBase.
+ */
+template <typename DURATIONTYPE>
+class DurationBase
+{
+    int64_t usec_; /**< Internal storage of the duration value in microseconds. */
```
pavel-kirienko: We should switch to `std::int*` instead of `uavcan::int` seeing as we can rely on the standard library now.

libuavcan/include/uavcan/time.hpp
```diff
@@ -0,0 +1,166 @@
...
+{
+/**
+ * Protected base class for duration values.
+ *
+ * @tparam DURATIONTYPE  The datatype returned when retrieving durations from
+ *                       realizations of this base class. This type must be
+ *                       exactly 8 bytes in size and must support certain concepts
+ *                       implied by DurationBase.
+ */
+template <typename DURATIONTYPE>
```
pavel-kirienko: Bad naming (also against the C++ Core Guidelines), should be PascalCase.

libuavcan/include/uavcan/time.hpp
```diff
@@ -0,0 +1,166 @@
...
+    {
+        return fromUSec(ms * 1000);
+    }
+
+    int64_t toUSec() const
+    {
+        return usec_;
+    }
+
+    int64_t toMSec() const
```
pavel-kirienko: I am mildly irked by the incompatibility of this naming convention with metric prefixes. SI and metric prefixes are no laughing matter.

Shall we rename these into `toMillisecond()`, `toMicrosecond()`, and so on?

libuavcan/include/uavcan/util/math.hpp
```diff
@@ -0,0 +1,227 @@
...
+ * Saturated addition where the RHS is assumed to be either a positive value or, if negative, a signal to
+ * add "max + 1" to the lhs.
+ *
+ * This is part of the generic, promotionless, saturation math provided by this header. Ideally this would
+ * be omitted for most platforms where saturation operations were available.
+ */
+template <
+    class SIGNED_TYPE,
+    typename std::enable_if<std::is_integral<SIGNED_TYPE>::value && std::is_signed<SIGNED_TYPE>::value, int>::type = 0>
+SIGNED_TYPE _saturating_add_d(SIGNED_TYPE left, SIGNED_TYPE right)
```
pavel-kirienko: I hope you appreciate how I am purposefully saying nothing about bad naming here. I am writing this to say that this function can probably be made constexpr. I am saying "probably" because constexpr functions are somewhat limited in C++11, and I'm not that much familiar with those limitations since I'm mostly dealing with C++17 here these days.

requirements.txt
```diff
@@ -0,0 +1,12 @@
+# Note that the cmake build will consume this for you if you have virtualenv
+# installed. If not you'll need to manually setup the correct python3 environment
+# and install these requirements.
+
+# libuavcan has a single python dependency although pydsdlgen itself will
+# have one or more dependencies including (but not necessarily limited to)
+# pydsdl
+
+pydsdlgen >= 1.0
```
pavel-kirienko: Did you mean `pydsdlgen < 2.0`?

pavel-kirienko:
(You may have noticed that I am using this PR as an opportunity to criticize some of my own code that I wrote years ago)

.clang-format
```diff
@@ -1,24 +1,87 @@
 ﻿---
```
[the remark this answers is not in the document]
thirtytwobits: I merged a lot of that one in but some of the directives are only available in a much newer version of clang-format. I want to support "state-of-the-art" tooling without requiring cutting edge versions (hense Ubuntu 18 LTS, cmake 3.10, arm-none-eabi-gcc 5.4, etc).

README.md
```diff
@@ -29,76 +32,91 @@ The reason we opted for a complete rewrite is that so much is changing. Between
...
+Contributors, please follow the [Zubax C++ Coding Conventions](https://kb.zubax.com/x/84Ah) and always use `clang-format` when authoring or modifying files (the build scripts will enforce but not apply the rules in .clang-format).
+
+### Pull-Request Checklist
+
+Reviewers, please check the following items when reviewing a pull-request:
+
+> **NOTE:** This is just the start of this checklist. Expect it to grow and get refined as this project matures.
+
+1. **correctness**
+    * Is the code correct.
```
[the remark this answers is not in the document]
thirtytwobits: In what way? That I have a checklist or in the contents of the checklist? Checklists are a best-practice for safety critical software development. That's why I'm developing this one for libuavcan.

cmake/compiler_flag_sets/cortex-m4-fpv4-sp-d16-nosys.cmake
thirtytwobits: The vendor code and gtest are written pretty loosely. I'm not as concerned about this given that we cover our code with stricter flags for the native builds.

cmake/compiler_flag_sets/native.cmake
thirtytwobits: Same question here as above?

requirements.txt
thirtytwobits: Not exactly. pydsdlgen < 1.0 won't work with this.  I'll change this to:

`pydsdlgen >= 1.0, < 2.0`

pavel-kirienko:
> Switched to using CMake external projects for our external dependencies (dsdl types and googletest). This means that after a clone you only have to mkdir build; cd build, cmake .. to get a full set of working makefiles and all our dependent sources.

This is cool but there's more to life than CMake.

Could you please briefly describe in a few sentences how would one integrate libuavcan v2.0 into a Make-based build process? I am picturing this:

- embed the libuavcan header files into the build tree (submodule, subtree, or copy-paste)
- `#include <uavcan/uavcan.hpp>`
- embed the DSDL source files into the build tree
- `pip install pydsdlgen`
- invoke pydsdlgenj (truly rolls off the tongue) from the Makefile

Am I missing anything here? Please let's not get unnecessarily consumed by CMake.

cmake/compiler_flag_sets/cortex-m4-fpv4-sp-d16-nosys.cmake
pavel-kirienko: I am questioning the necessity of `-Wno-missing-field-initializers`. Is there any third-party code which has missing field initializers, or is this merely a copy-pasting mistake?

README.md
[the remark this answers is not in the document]
pavel-kirienko: Okay, I see that this item is only needed for completeness and consistency. Please disregard this thread.

libuavcan/include/uavcan/util/math.hpp
thirtytwobits: I can make some of these constexpr by rewriting them using a ternary operator. For the ones that cannot be rewritten like this should I add, perhaps:
```
future_constexpr 
```
which will be constexpr when compiling using c++14 or greater? I'm not sure how much effort we want to put into forward optimizations given that we'll need to add additional builds for each dialect we want to support.

libuavcan/include/uavcan/time.hpp
thirtytwobits: sure! Will change.

libuavcan/include/uavcan/util/math.hpp
pavel-kirienko: Let us not trade correctness and robustness for resource utilization.

libuavcan/include/uavcan/time.hpp
thirtytwobits: Yep. This is a good convention; having only C macros appear ALL_CAPS. I'll follow from now on.

cmake/modules/clang-format-check.py
thirtytwobits: `-fallback-style=none` is something I should be doing but otherwise I like my solution which does not modify the source. 

No sooner that I had written this then I was made aware of a better OSS version (https://github.com/Sarcasm/run-clang-format). But mine works and I wrote it (god-dammit!).

libuavcan/include/uavcan/time.hpp
thirtytwobits: (I'm confused) int64_t is from the standard library?

thirtytwobits:
For your cmake comment: your bullet points are correct; that's basically what you need to do as a library consumer. I think there's a lot more user-facing documentation to be produced along this line in the future. I could also add a Makefile integration section below the section on [pydsdlgen cmake integration](https://pydsdlgen.readthedocs.io/en/latest/#cmake-integration) (you can pronounce it /pie-d'ess-delshen/ if you want to claim it's French).

libuavcan/include/uavcan/time.hpp
thirtytwobits: This confuses me. int64_t is from the standard library. Do you mean I should use std::? That's probably correct.

libuavcan/include/uavcan/porting/clock.hpp
thirtytwobits: yep

libuavcan/include/test/ptest_time.hpp
thirtytwobits: okay. libuavcan

libuavcan/include/uavcan/bus/can.hpp
thirtytwobits: Okay but there's one thing in the Zubax convention that bugs me. I think that only types and namespaces should start with a capital letter. For example:

```
class Widget
{
};

// Zubax
constexpr Widget FooWidget;

// Zubax + scott
constexpr Widget fooWidget;
```

Any chance I could convince you of this?

libuavcan/include/uavcan/bus/can.hpp
thirtytwobits: okay. will promote to namespace.

cmake/compiler_flag_sets/native.cmake
thirtytwobits: I think so. The way cmake works is these are keyed on the project's supported languages so you need flags for each language you declare.

cmake/compiler_flag_sets/cortex-m4-fpv4-sp-d16-nosys.cmake
thirtytwobits: Ah. Yes, we need it to build gtest for ontarget but the native one is a copy and paste error. Will remove.

cmake/compiler_flag_sets/native.cmake
thirtytwobits: changing to `-Wmissing-field-initializers`

thirtytwobits:
I'd like to merge this so I can finish hooking up CI. Can we agree this change isn't toxic and then we can continue the review with follow-up pull requests?

libuavcan/include/uavcan/time.hpp
pavel-kirienko: Yes, we must use `std::int64_t` because the C++ standard does not seem to define `::int64_t`.

thirtytwobits: Okay. This is already changed in the latest version of the PR.

pavel-kirienko: This reminds me of one spectacular performance issue that I uncovered in PX4 some years ago which was caused by the usage of C standard library functions instead of `std::`: https://github.com/PX4/Firmware/pull/6829#issuecomment-286920967

thirtytwobits: Interestingly enough, this is why we have to build gtest using `--gnu++11`

libuavcan/include/uavcan/bus/can.hpp
pavel-kirienko: I dreaded having this conversation. Do we really want to define a new set of coding conventions for libuavcan?

The general policy is that immutable stuff starts with a capital letter.

cmake/compiler_flag_sets/native.cmake
pavel-kirienko: My point is that we don't really have any C code here, do we?

libuavcan/include/uavcan/bus/can.hpp
thirtytwobits: image

pavel-kirienko:
> saturating add/subtract supporting all stdint types without compiler intrinsics

I forgot to ask: what do we need this for?

cmake/compiler_flag_sets/native.cmake
thirtytwobits: ah. okay. will remove this from the native set. thanks.

thirtytwobits:
I needed it for the duration math and then I went down a hole implementing it. I'm assuming it will become important for supporting saturated integer types in dsdl?

pavel-kirienko:
> I'm assuming it will become important for supporting saturated integer types in dsdl?

Not sure. Maybe not, unless you have a different approach to their support than what I have in mind. Basically, saturation is performed during serialization once per field, unless the field is of a standard bit length (8, 16, 32, 64) in which case it requires no additional handling.

thirtytwobits:
Crap, I forgot about #187

One more push with the removal of the C-only flags in native and changing the nascent "porting" namespace to be "platform"

PR #218, opened 2019-05-16
switching to nunuvut

thirtytwobits:


thirtytwobits:
I'm going to override this review since this is only a rename from pydsdlgen to nunuvut.

pavel-kirienko:
You seem to have misspelled it as `nunuvut` in the README (and in the issue title).

thirtytwobits:
and this is why one should never skip reviews :/

PR #219, opened 2019-05-17
fixing typo (nunavut)

thirtytwobits:
(sorry)


mhkabir:
(what does this word mean)

pavel-kirienko:
It means "a code generation library for DSDL". They also named a province in Canada after it: https://en.wikipedia.org/wiki/Nunavut

Issue #220, opened 2019-05-18
Rename "uavcan" namespace to "libuavcan"


thirtytwobits:
To prevent confusion and name collisions with generated DSDL types.

PR #221, opened 2019-05-20
Uavcan v1.0

thirtytwobits:
1. Adding python script to actually fail builds if on-target tests fail in the pipeline.
2. Issue #220 : renaming all the things to libuavcan – This looks like a big CR but it's only renames and documentation fixes.


test/native/test_transport_bus_can.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "libuavcan/uavcan.hpp"
+#include "libuavcan/bus/can.hpp"
+#include "gtest/gtest.h"
+
+namespace libuavcan
+{
+/**
+ * Verify the immutable properties of CAN FD
+ */
+TEST(CanBusTest, TypeFd)
```
pavel-kirienko:
```suggestion
TEST(CanBusTest, TypeFD)
```

test/native/test_transport_bus_can.cpp
```diff
@@ -0,0 +1,29 @@
...
+{
+    ASSERT_EQ(64U, bus::CAN::TypeFd::MaxFrameSizeBytes);
```
pavel-kirienko:
```suggestion
    ASSERT_EQ(64U, bus::CAN::TypeFD::MaxFrameSizeBytes);
```

test/native/test_build_config.cpp
```diff
@@ -4,14 +4,14 @@
...
+#include "libuavcan/uavcan.hpp"
 #include "gtest/gtest.h"
 
 /**
  * Test that the UAVCAN_VERSION_MAJOR number is expected.
  */
 TEST(BuildConfigTest, Version) {
-    ASSERT_EQ(2, UAVCAN_LIBRARY_VERSION_MAJOR) << "These tests are designed for version 2 of libuavcan" << std::endl;
+    ASSERT_EQ(2, LIBUAVCAN_VERSION_MAJOR) << "These tests are designed for version 2 of libuavcan" << std::endl;
     ASSERT_EQ(1, UAVCAN_SPECIFICATION_VERSION_MAJOR) << "These tests are designed for version 1 of the UAVCAN protocol" << std::endl;
```
pavel-kirienko: I am wondering if we should rename `UAVCAN_SPECIFICATION_VERSION_MAJOR` into `LIBUAVCAN_SPECIFICATION_VERSION_MAJOR`? One might object saying that there is no such thing as "libuavcan specification" though. But the upside of this change is that we're keeping libuavcan-related things properly prefixed.

test/native/test_build_config.cpp
```diff
@@ -4,14 +4,14 @@
  * Unit tests of the build_config.hpp header.
  */
 
-#include "uavcan/uavcan.hpp"
+#include "libuavcan/uavcan.hpp"
```
pavel-kirienko: Perhaps the main header should be renamed, too? `libuavcan/libuavcan.hpp`?

libuavcan/include/libuavcan/util/math.hpp
```diff
@@ -6,13 +6,13 @@
...
-#define UAVCAN_UTIL_MATH_HPP_INCLUDED
+#ifndef LIBUAVCAN_UTIL_MATH_HPP_INCLUDED
+#define LIBUAVCAN_UTIL_MATH_HPP_INCLUDED
 
-#include "uavcan/uavcan.hpp"
+#include "libuavcan/uavcan.hpp"
 #include <cstdlib>
 #include <cmath>
-#include "uavcan/introspection.hpp"
+#include "libuavcan/introspection.hpp"
```
pavel-kirienko: This note is not related to this PR but it seems important: you seem to be assuming here that the included path is located inside one of the include look-up directories since it's not reachable relatively. Quotes should be replaced with angle brackets (here and elsewhere).

test/native/test_build_config.cpp
thirtytwobits: I thought about that but I *think* this is valid as: "include uavcan from libuavcan." Put another way: if libuavcan had a different name, say "pavel" then this would be `#include "pavel/uavcan.hpp"` or "include uavcan from the pavel library". 

I dunno. What do you think?

libuavcan/include/libuavcan/util/math.hpp
thirtytwobits: <> first checks system paths whereas "" first checks local paths then system. The latter seems more correct for a header-only library since we do not expect it to be available on a system.

test/native/test_transport_bus_can.cpp
thirtytwobits: There's a few more placed I did this. Let be do a s&r and update the PR.

test/native/test_build_config.cpp
thirtytwobits: `LIBUAVCAN_UAVCAN_SPECIFICATION_VERSION_MAJOR`

It's long but it's correct.

test/native/test_transport_bus_can.cpp
thirtytwobits: There's a few more placed I did this. Let be do a s&r and update the PR.

test/native/test_build_config.cpp
pavel-kirienko: I'm thinking that the main header should be eponymous with the namespace of the library. But I don't have any strong preferences here either way.

libuavcan/include/libuavcan/util/math.hpp
pavel-kirienko: I see,

test/native/test_build_config.cpp
pavel-kirienko: > But I don't have any strong preferences here either way.

This is not to say that I am not inclined to prefer `libuavcan/libuavcan.hpp`. I am going to mark this as approved nevertheless but unless you disagree it's best to rename it.

Issue #101, opened 2017-04-02
Add explicit support for C++14 and newer


pavel-kirienko:
Currently the library cannot be compiled with C++14 and newer unless `UAVCAN_CPP_VERSION` is overridden.

antoinealb:
Bumping this as we are facing this issue when trying to build our project using C++14. Is there anything blocking support for C++14 or is it just a matter of adding correct detection logic to `build_config.hpp`?

pavel-kirienko:
Yes the change would be very trivial as you described. Or you could just define `UAVCAN_CPP_VERSION=UAVCAN_CPP11`.

thirtytwobits:
I think we need to fix up the top level CMakeLists.txt. I'll post a PR to see what we think of my plan.

thirtytwobits:
Closing as we allow it now. Issue #228 tracks remaining issues.

PR #227, opened 2019-05-27
Issue #101 allowing c++14

thirtytwobits:
This should allow compiling for c++14.
Caveat Emptor: We have not verified that libuavcan actually works using c++14 or 17. Compiles != works.


libuavcan/include/uavcan/build_config.hpp
```diff
@@ -25,6 +25,8 @@
  * standard than used by the compiler, in which case this symbol can be overridden manually via
  * compiler flags.
  */
+#define UAVCAN_CPP17    2017
+#define UAVCAN_CPP14    2014
 #define UAVCAN_CPP11    2011
 #define UAVCAN_CPP03    2003
```
pavel-kirienko: The conditional compilation expressions below this line need updating, too. Just fixing CMake scripts won't do because many people don't use CMake at all.

thirtytwobits: I'm unsure what you are looking for here. I have it so if you do not define `UAVCAN_CPP_VERSION` we error out if you try to use > C++11 since we haven't tested c++14 or higher. If you want to force it you can define `UAVCAN_CPP_VERSION=2014` and we'll let you do this. This change is just to allow you to build using the provided scripts with this override.

Do you want to actually support 14 by default? If so what tests would we need to verify this works? Specifically, I don't have any ability to test against px4 builds or verify on pixhawk hardware so I'm reticent to make such a change.

pavel-kirienko: I misunderstood your intent, sorry. This seems fine then.

@antoinealb does this work for you?

antoinealb: Well personally it does not really fix my issue, as I am not (yet) using the CMake build.

Plus if we add something to the build system that lets you override the library version detection to allow C++14, you might as well enable C++14, can't you ?

> Do you want to actually support 14 by default? If so what tests would we need to verify this works? Specifically, I don't have any ability to test against px4 builds or verify on pixhawk hardware so I'm reticent to make such a change.

Well I am assuming that for now either px4 overrides UAVCAN's version detection, or that they do not compile in C++14. In both cases they should not be affected by additional support for c++14 by UAVCAN.

I am not sure if we need anything else than the current unit tests and build though. C++14 is C++11 with extensions AFAIK, and it should not break anything. We can keep the CI on C++11 so that we do not accidentally use newer features.

thirtytwobits: But for your build @antoinealb, all you have to do is `-DUAVCAN_CPP_VERSION=2014` and, with this change, the tests and library will build and run correctly. All we are saying by not automatically allowing this is that we haven't tested c++14 at all so you'll need to set an override flag to acknowledge this. Does the Nuttx build even support c++14? I am aware of some major RTOSes that are still stuck on c++11 so we're being careful to avoid depending on it.

thirtytwobits: I'm inclined to merge this issue but I'll open a new issue and tag @antoinealb on it to track modifying build_config.cpp to automatically allow c++14.

Issue #222, opened 2019-05-21
Port driver layer to uavcan v1


thirtytwobits:
UAVCAN v1 will not have anything named "driver" in it so we want to move things that were in v0 "driver" into either platform or transport/bus.

pavel-kirienko:
FWIW in Pyuavcan, I've come up with the following layer structure for the network abstractions:

(application) <--> Node <--> Presentation <--> Transport <--> Media <--> (hardware)

Where "Presentation" implements the transport-agnostic communication logic (such as pub/sub, calls, DSDL serialization), and different transports implement an abstract interface `Transport`. Each transport implementation has an abstract `Media` interface specific for that transport which abstracts the implementation details for different types of backends, such as SocketCAN or SLCAN for the CAN transport, or the Berkeley sockets for the experimental UDP transport.

(redundant transports are supported through a special type of transport which aggregates several regular transports together according to the logic described in the spec)

The term "media" seems to match the usage well, so perhaps you might want to consider its use for libuavcan, too.

PR #237, opened 2019-05-30
Issue #222: minor progress.

thirtytwobits:
1. Improvments to docs
2. Shoring up namspacing for libuavcan validation suite.
3. Start of clock unit test.


thirtytwobits:
The buildkite build is failing because of the branch filters I had to add to ensure that master doesn't get built using buildkite. We're going to have to require that PRs come from branches with "uavcan-v1" somewhere in the branch name to support this.

thirtytwobits:
So, I'm asking for a review and if there are changes I'll rename the branch otherwise we'll override the PR build this one time.

PR #240, opened 2019-05-30
Uavcan v1/issue/230

thirtytwobits:
This is ready for review. now.


thirtytwobits:
Yeah. This won't work. Gotta tar up the html first.

pavel-kirienko:
Scott, would it be possible for me to opt out of reviewing pull requests that are not concerned with the general architecture or the runtime components of the library?

thirtytwobits:
> Scott, would it be possible for me to opt out of reviewing pull requests that are not concerned with the general architecture or the runtime components of the library?

Sure. I'll be sure to keep infrastructure changes separate from meaningful changes.

PR #241, opened 2019-06-01
Progress on duration and time.

thirtytwobits:
1. namespacing changes.
2. Adding support for unsigned +/- signed saturation math
3. Enforcing duration and time concepts.
4. Removing multiplication operations (we don't support saturation.
multiplication)

TODO:
finish unit tests for Monotonic time and duration.
Add UTC time abstraction.
Add media layer clock interface.


thirtytwobits:
I'll work on the on-target failure. My guess is this is a problem with gtest itself.

README.md
```diff
@@ -54,7 +54,7 @@ The reason we opted for a complete rewrite is that so much is changing. Between
 The following list of standardized* test environments will be used to validate the libuavcan implementation**:
 
 1. **Posix** - We will produce examples that can run on top of SocketCAN on Ubuntu 18 or newer. While expect that these examples will be generally compatible with other common linux distros or posix compliant operating systems (that also support SocketCAN) we will compile and test the examples using Ubuntu 18 as part of our CI build.
-1. **Bare-metal on NXP s32K146 devkit** - We expect to produce examples and tests that run on the s32K146 MCU populated on the standard s32146 evaluation board available from NXP. This is the primary test fixture for the project and will be used as the basis for specifying on-target test rigs.
+1. **Bare-metal on NXP s32K148 devkit** - We expect to produce examples and tests that run on the s32K148 MCU populated on the standard s32148 evaluation board available from NXP. This is the primary test fixture for the project and will be used as the basis for specifying on-target test rigs.
```
pavel-kirienko:
```suggestion
1. **Bare-metal on NXP S32K148 devkit** - We expect to produce examples and tests that run on the S32K148 MCU populated on the standard S32K148 evaluation board available from NXP. This is the primary test fixture for the project and will be used as the basis for specifying on-target test rigs.
```

libuavcan/include/libuavcan/time.hpp
```diff
@@ -14,153 +15,252 @@
...
+ * Protected base class for time values.
+ *
+ * @tparam Type          The type of the derived class. This must be an empty type.
+ *                       All storage will be provided by this base class.
+ * @tparam DType         The type of duration used for this time type. Time is concrete and duration
+ *                       is relative.
+ * @tparam USecT         The datatype returned when retrieving time from
+ *                       realizations of this base class. This type must be unsigned.
+ */
+template <typename Type, typename DType, typename USecT = uint64_t>
```
pavel-kirienko:
```suggestion
template <typename Type, typename DType, typename USecT = std::uint64_t>
```

(btw the SonarQube checker seems to have a rule for this)

libuavcan/include/libuavcan/time.hpp
```diff
@@ -14,153 +15,252 @@
...
+class Base
+{
+    USecT usec_;
```
pavel-kirienko:
```suggestion
    USecT usec_ = 0;
```

HIC++ 12.4.2, C++CG C.45

libuavcan/include/libuavcan/time.hpp
```diff
@@ -14,153 +15,252 @@
...
+
+protected:
+    ~Base() {}
+
+    Base()
+        : usec_(0)
```
pavel-kirienko:
```suggestion
```
HIC++ 12.4.2, C++CG C.45

libuavcan/include/libuavcan/time.hpp
```diff
@@ -14,153 +15,252 @@
...
+ * @tparam USecT         The datatype returned when retrieving durations from
+ *                       realizations of this base class. This type must be signed.
  */
-template <typename DurationType>
-class DurationBase
+template <typename Type, typename USecT = std::int64_t>
+class Base
 {
-    std::int64_t usec_; /**< Internal storage of the duration value in microseconds. */
+    USecT usec_; /**< Internal storage of the duration value in microseconds. */
```
pavel-kirienko:
```suggestion
    USecT usec_ = 0; /**< Internal storage of the duration value in microseconds. */
```
HIC++ 12.4.2, C++CG C.45

libuavcan/include/libuavcan/time.hpp
```diff
@@ -14,153 +15,252 @@
...
 protected:
     /**
      * Non-virtual destructor.
      */
-    ~DurationBase() = default;
+    ~Base() = default;
 
-    DurationBase()
+    Base()
         : usec_(0)
```
pavel-kirienko:
```suggestion
```
HIC++ 12.4.2, C++CG C.45

libuavcan/include/libuavcan/util/math.hpp
```diff
@@ -192,6 +192,28 @@ constexpr UNSIGNED_TYPE saturating_sub(UNSIGNED_TYPE left, UNSIGNED_TYPE right)
     return (right > left) ? std::numeric_limits<UNSIGNED_TYPE>::min() : static_cast<UNSIGNED_TYPE>(left - right);
 }
 
+template <class LHS_TYPE,
+          class RHS_TYPE,
```
pavel-kirienko: LhsType, RhsType

libuavcan/include/libuavcan/time.hpp
```diff
@@ -14,153 +15,252 @@
...
     {
-        static_assert(sizeof(DurationType) == 8, "DurationType must be 64 bits wide.");
+        static_assert(sizeof(Type) == sizeof(USecT),
+                      "The duratiotn abstraction must be the same size as the underlying duration type.");
```
pavel-kirienko:
```suggestion
                      "The duration abstraction must be the same size as the underlying duration type.");
```

libuavcan/include/libuavcan/time.hpp
thirtytwobits: Actually, by C.21 and because we want to define the default ctor as protected I need to implement all [default operations](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#cctor-constructors-assignments-and-destructors)

libuavcan/include/libuavcan/time.hpp
thirtytwobits: Same comment as above: per c.21 will define all default operations.

libuavcan/include/libuavcan/time.hpp
thirtytwobits: fixed this in my branch. Update will be provided shortly...

libuavcan/include/libuavcan/time.hpp
thirtytwobits: Same comment as above: per c.21 will define all default operations.

libuavcan/include/libuavcan/time.hpp
thirtytwobits: See comment below: will define all defaults.

libuavcan/include/libuavcan/time.hpp
pavel-kirienko: But the constructor can be reduced to `Base() = default;`, can it not? There is no point putting static asserts there. My underlying concern is that the initializer list is used where it could be avoided. Or perhaps I am overreacting and this issue is so minor it shouldn't even be discussed?

thirtytwobits: Actually, my compiler refuses to allow the static_assert outside of the ctor.

```
libuavcan/include/libuavcan/time.hpp:34:19: error: invalid application of 'sizeof' to an incomplete type 'libuavcan::duration::Monotonic'
    static_assert(sizeof(Type) == sizeof(USecT),
```

pavel-kirienko: Okay, it makes sense. What do we do with the field initializers? Not just here but in general in similar cases.

thirtytwobits: I would seem, according to C.21, that we must provide them all if we provide any. The compilers will complain about missing or out of order initializers. We should also get Coverity scans running again which are really good at catching this sort of thing.

PR #245, opened 2019-06-08
Progress on the media layer

thirtytwobits:
1. some docs cleanup.
2. completed the time.hpp unit tests.
3. Added timestamps to the can frame
4. Added formal way to change the microsecond types used by the library.


libuavcan/include/libuavcan/build_config.hpp
```diff
@@ -62,6 +66,14 @@
 #    define LIBUAVCAN_EXPORT
 #endif
 
+/**
+ * Allows selecting the size of the microsecond timestamp type used by libuavcan.
+ * Valid values are 8, 4, and 2.
+ */
+#ifndef LIBUAVCAN_MICROSECOND_SIZE_BYTES
+#    define LIBUAVCAN_MICROSECOND_SIZE_BYTES 8
```
pavel-kirienko: Why 2 and 4 are valid? The point of 64-bit integers is that they never overflow, removing an (ostensibly) common source of errors in time handling logic where the value wraps over to zero. Allowing smaller integers would mean that the code will have to account for overflows, bringing additional complexity and possibly jeopardizing correctness. Do we really want this, or am I misunderstanding the intent? What do we gain by allowing the variability here?

libuavcan/include/libuavcan/time.hpp
```diff
@@ -144,7 +181,14 @@ class Base
...
     {
-        return fromMicrosecond(-usec_);
+        if (usec_ == std::numeric_limits<USecT>::max())
+        {
+            return fromMicrosecond(std::numeric_limits<USecT>::min() + 1);
+        }
+        else
+        {
+            return fromMicrosecond(-usec_);
+        }
```
pavel-kirienko: Given that for a two's complement signed integer `abs(min) == max + 1`, does this code handle the case where `usec_ == std::numeric_limits<USecT>::min()` correctly? I couldn't find this case covered in the unit tests but maybe I just missed it.

libuavcan/include/libuavcan/build_config.hpp
thirtytwobits: I had an argument for why I did this but I don't feel like defending it right now. I'll just  remove this and if it comes up in the real world we can revisit the argument then.

libuavcan/include/libuavcan/time.hpp
thirtytwobits: erp. Fixed this. Thanks.

thirtytwobits:
You can ignore the last change as it is just infrastructure stuff you didn't want to get involved in. Getting coverage reports for  C++11  is  HARD, it turns out. Jeebus.

thirtytwobits:
> Oops, I just noticed that the microsecond type width is still configurable.

But not really? It's configurable if you use the Base type but not for how the library will use Monotonic types.

pavel-kirienko:
> But not really? It's configurable if you use the Base type but not for how the library will use Monotonic types.

I mean why do we need the `LIBUAVCAN_MICROSECOND_SIZE_BYTES` build config parameter?

thirtytwobits:
> LIBUAVCAN_MICROSECOND_SIZE_BYTES

ack! Crap. Sorry, fixing...

PR #248, opened 2019-06-09
DO NOT MERGE YET

thirtytwobits:
This is a design review  PR. Please comment on the design and we'll
evolve it before it's ready for  merging.

### DIFF
--- a/libuavcan/include/libuavcan/transport/media/channels.hpp
+++ b/libuavcan/include/libuavcan/transport/media/channels.hpp
@@ -1,0 +1,195 @@
+/*
+ * Copyright (C) 2014 Pavel Kirienko <pavel.kirienko@gmail.com>
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ */
+/** @file */
+
+#ifndef LIBUAVCAN_TRANSPORT_MEDIA_CHANNELS_HPP_INCLUDED
+#define LIBUAVCAN_TRANSPORT_MEDIA_CHANNELS_HPP_INCLUDED
+
+#include <array>
+#include <algorithm>
+
+#include "libuavcan/libuavcan.hpp"
+#include "libuavcan/introspection.hpp"
+#include "libuavcan/time.hpp"
+
+namespace libuavcan
+{
+/**
+ * @namespace transport
+ * Contains transport-specific types and namespaces.
+ */
+namespace transport
+{
+/**
+ * @namespace media
+ * Contains media-specific types and namespaces.
+ *
+ * See the @ref MediaDevGuide for details
+ * on porting the media layer to a given platform.
+ */
+namespace media
+{
+/**
+ * Filter config struct.
+ * TODO: document and examples
+ */
+struct LIBUAVCAN_EXPORT FilterConfig
+{
+    std::uint32_t id;
+    std::uint32_t mask;
+
+    bool operator==(const FilterConfig& rhs) const
+    {
+        return rhs.id == id && rhs.mask == mask;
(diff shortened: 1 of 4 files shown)


libuavcan/include/libuavcan/transport/media/channels.hpp
```diff
@@ -0,0 +1,195 @@
...
+};
+
+/**
+ * Single non-blocking connection to a UAVCAN bus with isolated rx and tx queues. While channels
+ * logically operate independently they may share physical peripherals for some platforms.
+ *
+ * @tparm  FrameType    The media-specific frame type to be exchanged on this channel.
+ */
+template <typename FrameType>
+class LIBUAVCAN_EXPORT Channel
```
pavel-kirienko: Usually, OS designers call these things "interfaces". Do we not want to follow the same convention?

thirtytwobits: Channel works for both low-level integrations, where there is a 1:1 relationship between this object and a hardware peripheral, and for high-level integrations where the "interface" is logical/virtual. Specifically, socketCAN makes more sense in this naming scheme.

libuavcan/include/libuavcan/transport/media/channels.hpp
```diff
@@ -0,0 +1,195 @@
...
+    std::uint32_t mask;
+
+    bool operator==(const FilterConfig& rhs) const
+    {
+        return rhs.id == id && rhs.mask == mask;
+    }
+
+    FilterConfig()
+        : id(0)
+        , mask(0)
```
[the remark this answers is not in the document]
thirtytwobits: ?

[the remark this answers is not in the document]
pavel-kirienko: Sorry, I should have been more verbose. What I was trying to communicate is that we should use field initializers instead of ctor init list here, per the rules we've discussed earlier.

libuavcan/include/libuavcan/transport/media/channels.hpp
pavel-kirienko: Okay.

thirtytwobits: ...except I now realize that the specification is very explicit about this terminology even requiring "interfaces" to have "indices" in order to implement the requirement that "interfaces with (sic) lower numerical index should be handled in the first order."

I'll change back to calling these interfaces.

thirtytwobits:
Closing this one. Will have a real PR up...uh...soon.

libuavcan/include/libuavcan/transport/media/channels.hpp
pavel-kirienko: > the specification is very explicit about this terminology even requiring "interfaces" to have "indices" in order to implement the requirement that "interfaces with (sic) lower numerical index should be handled in the first order."

The specification does that simply because one needs words to refer to things in English. I did not intend to require anybody to follow any particular naming convention or have an explicit notion or indices; instead, this should be interpreted abstractly at the behavioral level. Should we add a note somewhere at the beginning of the doc to prevent people from reading too much into the text?

Despite the above, I still think that "interface" makes more sense (marginally; for example, when you call `netstat -i` it says that `can0` is an "interface") but I am not against "channel" either.

pavel-kirienko: > The specification does that simply because one needs words to refer to things in English.

Wouldn't be a problem if the behavior was described through a more abstract mathematical model. No words = less chance for misinterpretation.

Issue #254, opened 2019-06-24
Infinite loop in the AVL tree


pavel-kirienko:
It has been reported that there is allegedly an infinite loop in the AVL tree:

https://forum.uavcan.org/t/libuavcan-cubemx-freertos-systemworkbenchforstm32/430/9?u=pavel.kirienko

dpslwk:
fix suggested by kolos450 in the linked thread got this working for me on an STM32L433 under FreeRTOS

pavel-kirienko:
For reference, this is the suggested fix copy-pasted from the forum verbatim:

> Hi. I've got the same issue. There is a typo in the appendToEndOf method:
> ```
> target = head->equal_keys;
> ```
> should be 
> ```
> target = target->equal_keys;
> ```
> The case happens when the driver is busy and all the frames produced from a single message are sent to a linked list.

@thirtytwobits should we just go ahead and fix this, seeing as @Zarkopafilis is unavailable?

pavel-kirienko:
Related: https://forum.uavcan.org/t/how-to-support-getnodeinfo-and-getdatatypeinfo-on-stm32/577

thirtytwobits:
I'll take a look

thirtytwobits:
All fixed. Thanks for the bug report and sorry for the slow response.

Issue #190, opened 2018-12-19
Enforcement of the coding conventions


pavel-kirienko:
The coding conventions are described here: https://kb.zubax.com/x/84Ah

The existing codebase has certain deviations which should be fixed in the rewrite; the two things that come to mind are the capitalization rules (`CANFrame` is correct, `CanFrame` is not) and reliance on the native types (`std::uint32_t` is safe to use, `unsigned` is not). Also, a lot of useless stuff can be safely thrown out now since we have agreed to abandon support for C++03.

We need some tool for enforcement of MISRA C++ rules. I Googled around a bit and found nothing except SciTools Understand, which I have a bit of experience with: https://scitools.com/features. Back in 2014 they offered a free license for open source projects, not sure if they still do. While the tool generally worked for me, it was rather unreliable and tended to produce nonsensical reports, especially when dealing with heavily templated code which libuavcan has much of. Seeing as we have no other options so far, we should probably give it a try.

Besides MISRA, we could well use more static analysis in general. We currently have Coverity Scan; perhaps we should also consider using the [Clang Static Analyzer](https://clang-analyzer.llvm.org/), which I heard good things of but never used myself.

pavel-kirienko:
@thirtytwobits SciTools bestowed one free license upon us. It is valid for one year, can be freely renewed afterward.

I asked them whether it is possible to set up their tool in a headless mode in a CI environment, but they ignored this question twice, so I take it as a "no".

thirtytwobits:
This task has become a bit too non-specific for v1. We now have enforcement of style and Issue #226 covers integrating clang-tidy but we do lack MISRA and/or HIC++ enforcement. Can I change this to read "Choose and enforce a C++ standard for safety critical software"?

pavel-kirienko:
I think I might have found just what we need. SonarQube has a cloud-based static analysis service named SonarCloud which is free for open source projects, and it supports a reasonable subset of MISRA (https://rules.sonarsource.com/cpp/tag/misra, https://rules.sonarsource.com/cpp/tag/misra-c++2008) among other rules. I have set it up for the new [PyUAVCAN rewrite](https://github.com/UAVCAN/pyuavcan/compare/436011bc7448c54f1ab13b7563e8b28bf1fc14a0..6507e1bec55f2d3299eb41affa346e17acdff55b) which is my first ever encounter with this service; currently I can't vouch for its C++ side so we need to try it. The list of rules at least looks very promising.

[...]

PR #265, opened 2019-08-04
Issue #254, fixing infinite loop in AVL tree

thirtytwobits:
This test reproduced the problem and verified that fixing the typo fixes the problem.


thirtytwobits:
I've removed the codacy integration. It's just not working. We'll have to look at an alternative that works better for native code.

pavel-kirienko:
> I've removed the codacy integration. It's just not working. We'll have to look at an alternative that works better for native code.

SonarQube looks promising https://github.com/UAVCAN/libuavcan/issues/190

Issue #235, opened 2019-05-29
Port transport layer to v1


thirtytwobits:
This is a placeholder. Hopefully I can break this down into several/many tasks before starting on the Alpha milestone.

pavel-kirienko:
Sub-tasks:
https://github.com/UAVCAN/libuavcan/issues/181
https://github.com/UAVCAN/libuavcan/issues/185
https://github.com/UAVCAN/libuavcan/issues/186

Issue #205, opened 2019-02-06
scalar_codec.hpp UAVCAN_ASSERT on big_endian processor


jacobschloss:
It looks like there is an assert for big endian processors based on portability concerns / lack of testing.

We are running libuavcan on a big endian processor (PowerPC e200z7, NXP MPC5777M), so will be able to provide testing. Once we confirm that it all works, can we remove [this assert](https://github.com/UAVCAN/libuavcan/blob/dfcdf22eda16ff06847976fd6c7f40671fc92eb5/libuavcan/include/uavcan/marshal/scalar_codec.hpp#L41) in convertByteOrder? We'll probably have it up and running this week.

pavel-kirienko:
Yes, please do that, and send us a pull request, too. Thanks!

PR #266, opened 2019-08-07
Progress on #235, adding unaligned copy.

thirtytwobits:
Adding a version of unaligned copy for little-endian platforms that
provides a single abstraction for vectorized unaligned copy. This
replaces the v0 BitStream + unalignedCopy abstractions making a v1
"bit stream" into simply an std::vector and a std::size_t bit_offset.

The abstraction allows the vector's word size to be parameterized which
could provide better optimizations where a given architecture can
perform single operations across multiple bytes instead of hard-coding
all 8-bit operations. The usefulness versus complexity of this
optimization will play out in subsequent pull-requests as we start
utilizing this copy operation in the generated dsdl types.


thirtytwobits:
Sorry. Didn't have time to fix the gcc errors. Will do that tomorrow but I wanted to get the code to you sooner for review.

libuavcan/include/libuavcan/platform/memory.hpp
```diff
@@ -304,6 +323,126 @@ class LIBUAVCAN_EXPORT PoolAllocator
     }
 };
 
+namespace
```
pavel-kirienko: While working on the PyUAVCAN serialization logic I noticed that the complexity of the bit copy algorithm (in terms of operations per byte copied) can be reduced by taking advantage of the fact that either the source or the destination are always aligned at one byte:

- https://github.com/UAVCAN/pyuavcan/blob/uavcan-v1.0/pyuavcan/dsdl/_serialized_representation/_serializer.py#L162-L172
- https://github.com/UAVCAN/pyuavcan/blob/uavcan-v1.0/pyuavcan/dsdl/_serialized_representation/_deserializer.py#L221-L252

More specifically, for N bytes to copy, an implementation that assumes that either the source or the destination are aligned performs N or N+1 iterations of the outer loop (depending on the alignment at the opposite end), compared to (approximating here) N...2N of the generic method.

libuavcan/include/libuavcan/platform/memory.hpp
```diff
@@ -304,6 +323,126 @@ class LIBUAVCAN_EXPORT PoolAllocator
...
+    *dst = DstType((*dst & ~write_mask) | (DstType((*src >> src_bit_offset) << dst_bit_offset) & write_mask));
+}
+}  // namespace
+
+/**
+ * This function implements a portable copy of unaligned bit arrays that avoids generation of
+ * unaligned instructions for a platform.
+ *
+ * Note that this implementation makes a strong assumption that the platform is
+ * little endian. This logic should completely fail on a big-endian system.
```
pavel-kirienko: Shouldn't this be guarded by a preprocessor check? If we could detect C++20 it could be a constexpr check https://en.cppreference.com/w/cpp/types/endian

People use libuavcan on big-endian machines, see #205, #210

libuavcan/include/libuavcan/platform/memory.hpp
thirtytwobits: Okay. Yeah. splitting this up into to methods simplifies the logic quite a bit. I'll also just hard-code uint8_t as the work type to make this endian agnostic.

libuavcan/include/libuavcan/platform/memory.hpp
thirtytwobits: added #267 but will also make this work on 8-bit words only. I assume we don't plan on supporting 4-bit processors? ;-)

pavel-kirienko: Hail standardization.

thirtytwobits:
Okay. I think this is good. Honestly I won't know for sure till I start using it. The next step is to start building the presentation layer which will utilize this.

libuavcan/include/libuavcan/platform/memory.hpp
```diff
@@ -304,6 +304,100 @@ class LIBUAVCAN_EXPORT PoolAllocator
...
+            current_byte      = offset_bits / 8U;
+            dst[current_byte] = static_cast<std::uint8_t>(src[current_byte] << bits_from_src_byte);
+            bits_copied += local_offset;
+        }
+        else
+        {
+            // we don't need to reevaluate the while condition.
+            break;
+        }
+    } while (true);
```
pavel-kirienko: Any reason to avoid `for (;;)`?

thirtytwobits: You propose we lie with the emptiness ; between ; the semicolons ; ; the void that questions all symbols ; that screams with the horror of its own, undefined nature ;::.;.?!|

Nay, sir! I embrace the life-affirming : (all things are true)

Cynic.

Just a preference. It's the exact same set of instructions ether way.

pavel-kirienko: Convincing. Is there also a pearl of unspoken wisdom about the preference for `do ... while (true)` instead of `while (true)` that I should know about?

thirtytwobits: nope. My weird psychology around this is just that, since we want the loop evaluation to go away, it's best to hide it at the bottom, so only human factors are involved. I'm also one of those people that put private stuff in classes at the bottom of the definition but I've had to change my ways for Zubax style!

pavel-kirienko: People following the Zubax style should be thankful for saving one `private` per class definition.

image

Issue #264, opened 2019-07-30
Move all cmake machinery under tst


thirtytwobits:
To reduce confusion we want to move all the cmake machinery we currently have to live under the test folder. At the top level we should provide something very basic (Makefile?) and documentation to help users integrate libuavcan and nnvg with their application.

pavel-kirienko:
> At the top level we should provide something very basic (Makefile?) and documentation to help users integrate libuavcan and nnvg with their application.

Maybe we should provide *nothing* on the top level except for documentation? I recall seeing users of IAR or Keil IDE not knowing what to do with a makefile. Examples for particular build systems could be moved somewhere into deeper chapters of the user documentation.

PR #292, opened 2020-02-22
issue #264: moving CMakeLists

thirtytwobits:


thirtytwobits:
> Also, is there any strong reason why `doc_source` is not `docs`?

because `docs` implies you can go there and read that stuff from source. I'm indicating that you need to build the docs from this source.

thirtytwobits:
> Ah. Finally.

Yeah, I'm rushing to get sonarqube all setup for Nuno and to cleanup some of this boring infrastructure stuff we don't want to waste his time with.

PR #305, opened 2020-03-26
Refactor Linux Driver

thirtytwobits:
I did a similar refactor downstream to support a posix-like RTOS
internally. I'm submitting this but with our RTOS-specific driver
removed to allow Peter to refactor his nuttx driver to share most
of its implementation with linux.


pavel-kirienko:
Why did you remove the LPC11C24 driver?

pavel-kirienko:
Actually, none of the platform drivers should be here anyway. I suggest moving this changeset over to https://github.com/UAVCAN/platform_specific_components/tree/legacy-v0/linux/libuavcan.

thirtytwobits:
> Why did you remove the LPC11C24 driver?

It seemed incredibly out of date. When's the last time anyone actually verified this worked? I can add it back but is it at all relevant at this point?

pavel-kirienko:
It is certainly not out of date and it is actually used in OpenGrab EPM v3, for example. If it's not being updated it doesn't mean that it's dead, it's just finished and there's nothing else to change.

pavel-kirienko:
Closing due to #307.

PR #306, opened 2020-03-28
PX4 legacy-v0 downstream changes

MaEtUgR:
Pr based on request here https://github.com/PX4/libuavcan/pull/3#issuecomment-605483566

### DIFF
--- a/libuavcan_drivers/stm32/driver/src/internal.hpp
+++ b/libuavcan_drivers/stm32/driver/src/internal.hpp
@@ -11,7 +11,7 @@
 #elif UAVCAN_STM32_NUTTX
 # include <nuttx/arch.h>
 # include <arch/board/board.h>
-# include <chip/stm32_tim.h>
+# include <stm32_tim.h>
 # include <syslog.h>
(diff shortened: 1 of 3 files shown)


libuavcan_drivers/stm32/driver/src/internal.hpp
```diff
@@ -11,7 +11,7 @@
 #elif UAVCAN_STM32_NUTTX
 # include <nuttx/arch.h>
 # include <arch/board/board.h>
-# include <chip/stm32_tim.h>
+#include <stm32_tim.h>
```
pavel-kirienko:
```suggestion
# include <stm32_tim.h>
```

MaEtUgR: fixed through rebase

pavel-kirienko:
The Kinetis submodule change is erroneous, I missed it in my review. The current changeset rolls the submodule back to an older revision.

I am going to accept this PR anyway because after it's taken in I will submit a yet another PR removing `libuavcan_drivers` completely (it's been announced long ago so hopefully everyone is prepared for that now).

PR #318, opened 2021-07-28
update pyuavcan lib version and name

PonomarevDA:
After updating `uavcan` python library, libuavcan_dsdlc stopped working with import error: `ImportError: No module named uavcan`.
In this PR I simply renamed `uavcan` lib to `pyuavcan_v0` and updated `pyuavcan` submodule because as I understand previous version had different dsdl path. Now it seems to be ok.


pavel-kirienko:
@coderkalyan does it look okay to you?

coderkalyan (approved):
LGTM. There don't seem to be any other references to the `uavcan` package.

pavel-kirienko (approved):

PR #320, opened 2022-12-02
Replace uavcan with cyphal terminology

aeriq:
As part of the rebrand to "Cyphal", replaced references to UAVCAN with the Cyphal equivalent terminology.


CONTRIBUTING.md
```diff
@@ -1,38 +1,38 @@
...
 
-v1.0 is a complete rewrite of libuavcan with the following, fundamental changes from v0:
+v1.0 is a complete rewrite of libcyphal with the following, fundamental changes from v0:
 
-1. libuavcan v1 is based on the [UAVCAN v1 specification](https://new.uavcan.org).
-2. libuavcan v1 requires C++11 or greater.
-3. libuavcan v1 favors idiomatic C++ over defining its own utilities and helpers.
-4. libuavcan v1 is a header-only library.
-5. libuavcan v1 holds itself to a higher quality standard and is designed for integration into high-integrity, embedded applications.
+1. libcyphal v1 is based on the [Cyphal v1 specification](https://new.opencyphal.org).
```
thirtytwobits, line 5:
```suggestion
1. libcyphal v1 is based on the [Cyphal v1 specification](https://opencyphal.org).
```

CONTRIBUTING.md
```diff
@@ -1,38 +1,38 @@
...
+The reason we opted for a complete rewrite is that so much is changing. Between the updates to the specification, the abandonment of C++98 support, switching to header-only, removing the drivers from the main repository, adding CAN FD support, etc, etc, etc; it was obvious that we'd be rewriting everything anyway. The good news is that v0 exists, is fully supported, and will be liberally copy-and-pasted from as this makes sense for v1. What we don't have is any git history tracing from v0 to v1 since this would be deceiving. Futhermore, the unit tests in libcyphal v0 are a bit of a mess so we'll be writing them in a way that is more sustainable.
```
thirtytwobits, line 17:
```suggestion
The reason we opted for a complete rewrite is that so much is changing. Between the updates to the specification, the abandonment of C++98 support, switching to header-only, removing the drivers from the main repository, adding CAN FD support, etc, etc, etc; it was obvious that we'd be rewriting everything anyway. The good news is that v0 exists, is fully supported, and will be liberally copy-and-pasted from as this makes sense for v1. What we don't have is any git history tracing from v0 to v1 since this would be deceiving. Furthermore, the unit tests in v0 are a bit of a mess so we'll be writing them in a way that is more sustainable.
```

thirtytwobits (approved):

thirtytwobits:
Folding this back into another commit I'm about to push that will include github actions...

PR #327, opened 2023-06-07
integrating nunavut

thirtytwobits:
I am so ... sick ... of ... cmake!

### DIFF
--- a/docs/examples/example_01_hello_world.cpp
+++ b/docs/examples/example_01_hello_world.cpp
@@ -20,22 +20,67 @@
 
 #include "example/foo_1_0.hpp"
 
-using byte = cetl::pf17::byte;
+/// We'll create something like this in Nunavut so I'm just providing some foreshadowing here:
+/// @tparam T
+template <typename T>
+struct dsdl_traits
+{
+    static constexpr std::size_t extent() noexcept
+    {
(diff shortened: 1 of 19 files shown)


docs/examples/example_01_hello_world.cpp
```diff
@@ -14,11 +14,28 @@
 
 #include <gtest/gtest.h>
 
+#include "cetl/pf17/sys/memory_resource.hpp"
+#include "cetl/pf17/byte.hpp"
+#include "cetl/variable_length_array.hpp"
+
+#include "example/foo_1_0.hpp"
```
lydia-at-amazon, line 21: Did you also have an example of including a regulated data type?

lydia-at-amazon (approved):

lydia-at-amazon:
Should we expect the checks to pass?

emrainey (approved):
Would like to see some usage w/ the common standard messages (like one or two) but that doesn't have to be here.

docs/examples/example_01_hello_world.cpp
thirtytwobits: Wow. It's like you two don't trust me ;-) Yah. Will add...

emrainey (approved):

PR #330, opened 2023-07-06
progress on Transport layer

thirtytwobits:
Bulk change over from previous design to pycyphal-like design. At this checkin we confirmed the network layer was functional on osx. There should be little difference, if any, for linux. The transport layer is 90% complete with the read and write runFor methods being the last integration point needed for an end-to-end test.

This is mostly a design review and I'd encourage you to checkout the change locally and review that way instead of trying to cut through the noise of the diff (which is more deleted files than added).


pavel-kirienko (changes requested):
As discussed at the weekly call yesterday, I think the design based on the generic `Status` type is not a good one as it attempts to solve the problem of execution status reporting using an overly generalized, rigid abstraction. One can construct more typesafe and expressive interfaces by using `expected` or `optional<MyErrorType>` ad-hoc rather than defining a universally applicable type.

The same applies to getFlags, makeFlags, and the large one-size-fits-all `ResultCode`. I think these are better removed in favor of more granular result types as it helps with a more orthogonal, less coupled design.

I also suggest avoiding excessive reliance on output arguments, like in `IPolymorphicType::queryType`, and using `expected` instead.

Not sure if the library needs to be involved in highly platform-specific activities such as the construction of new network interfaces. As I mentioned at the call, the `IContext` interface is a questionable design choice because it forces the coupling of largely independent entities -- transport layer implementations -- through itself in a rather artificial way. If the application doesn't want Cyphal/serial, it shouldn't need to stub out factory methods for it; likewise, addition of new transports should not break existing applications.

Another issue that is tightly related is that it is, I think, a mistake to apply the same lifetime management principles to CAN sockets and UDP sockets, as they serve entirely different purposes within their transport implementations. A CAN socket is a CAN interface; it is created once when the transport is initialized and remains alive until it is disposed of; there is no need to construct/destroy CAN sockets dynamically (but see below about hot-plugging). A UDP socket is a Cyphal/UDP port, a higher-level entity; ports need to be constructed and destroyed whenever the application needs to change the set of ports it is interested in (begin publishing, subscribe, begin providing an RPC-service, etc). Hence, I think the existing common socket abstraction is not a good one and I recommend removing it.

Regarding the interface hot-plugging: I think this concern can be easily moved outside of the scope of the library if we follow the `RedundantTransport` approach implemented in PyCyphal. The idea here is that there is a container object that implements `ITransport` and hosts a set of underlying `ITransport` instances. Whenever a transfer is emitted, it is sent through each of the underlying transports; when a transfer is received, it goes through a deduplicator. One desirable side-effect of this approach is that it allows you to change the transport composition underneath `RedundantTransport` dynamically, thus facilitating hot-plugging. One caveat here is that this mode of redundancy will exist side-by-side with the homogeneous redundancy support implemented in the lib*ards, but I don't see this as a significant issue and it is either way required to support heterogeneous redundancy. See https://pycyphal.readthedocs.io/en/stable/api/pycyphal.transport.redundant.html

The sessions of the UDP transport will need to be remodeled after the new libudpard API; `toMulticastAddress` will likely need to be removed. We can revisit this part after my branch of libudpard is finished (soon :soon:).

The `TransportStatistics` class probably should not remain polymorphic as I expect it's going to be awkward to use with our makeshift reflection support with IPolymorphicType. For now, I would leave this part out entirely, as it is low-priority and can be added later easily.

Also regarding reflection, the data specifier should be an algebraic type like `using DataSpecifier = cetl::variant<MessageDataSpecifier, ServiceDataSpecifier>` because the lack of proper RTTI makes certain conventional OOP design patterns suboptimal here. There probably are other cases where runtime-polymorphic types need to be replaced with algebraic alternatives.

The ITransport interface will need to be extended with some introspection getters to provide access to the set of existing input/output sessions, similar to the `input_sessions`/`output_sessions` properties in PyCyphal.

I think the `initialize` method is rightfully considered to be [kind of an anti-pattern](https://isocpp.org/wiki/faq/exceptions#ctors-can-throw) these days. I suggest we replace it with freestanding fallible factory functions that are friends of the classes they instantiate (the classes would have private constructors). This is effectively like having fallible constructors with return codes:

```c++
enum class Error;
class Foo final : public IFoo
{
public:
    // good stuff
private:
    friend std::expected<Foo*, Error> makeFoo(...);
    Foo();
};
inline std::expected<Foo*, Error> makeFoo(...);
```

The naming chosen for Service Provider/Consumer refers to SOA rather than RPC; a more appropriate term would be Server/Client. It is important to follow consistent naming throughout the project to avoid confusion.

>Balance between type-safety and SLoC

We're on the same page.

PR #333, opened 2023-12-31
Add the new design doc

pavel-kirienko:
This changeset may seem large but it is actually very simple:

- The source code is dropped. Bye bye source code.
- The design document `docs/design/readme.md` is added along with a pic. Please read the doc and admire the pic.

Happy new year.


docs/design/readme.md
```diff
@@ -1210,4 +1210,495 @@ Server instances are unique per service-ID; as such, the interface differs sligh
...
+                    // Shall it fail, the error is likely to be corrected during the next save().
+                    (void) rgy.set(reg_name, value);
+                }
+            }
+        }
+    }
+    return std::nullopt;
+}
+
+/// The register savior is the counterpart of load().
```
emrainey, line 1641: savior -> saver? commit-o-tron?

pavel-kirienko:
@emrainey this is the part of `std::any` that substitutes RTTI with an identical mechanism: https://github.com/gcc-mirror/gcc/blob/a56a693a74dd3bee71b1266b09dbd753694ace94/libstdc%2B%2B-v3/include/std/any#L530-L535.

docs/design/readme.md
pavel-kirienko: eh

pavel-kirienko:
TODO:

- [ ] Add non-typed versions of `Publisher`/`Subscriber`/`Client`/`Server` and corresponding getter methods to the `Presentation` controller to allow type-agnostic usage of the presentation layer.

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+
+```c++
+/// A simplified substitute for RTTI that allows querying the identity of types.
+/// Returns an object of an unspecified type for which operator== and operator!= are defined.
+/// Objects for the same type compare equal. Ordering is not defined.
+template <typename>
+auto getTypeID() noexcept
+{
+    static const struct {} placeholder{};
+    return static_cast<const void*>(&placeholder);
```
emrainey, line 55: Should return a `void const * const` since it's static. or just `void * const` if you are not trying to return a constant void but are trying to return a constant pointer.

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+    PublisherBase(PublisherBase&&)                 = default;
+    PublisherBase& operator=(const PublisherBase&) = delete;
+    PublisherBase& operator=(PublisherBase&&)      = default;
+    virtual ~PublisherBase() noexcept              = default;
+
+protected:
+    explicit PublisherBase(const SharedPtr<PublisherImpl>& impl);
+
+private:
+    SharedPtr<PublisherImpl> impl_;
```
thirtytwobits, line 967: Let's talk about `SharedPtr` here... Why does PublisherImpl have to be reference counted? Besides the total lack of control over an object's lifecycle and the enabling of very unsafe multi-threaded code, sharing pointers indiscriminately is normally just lazy and rarely required once the use case is properly analysed. If the requirement is reuse of dynamic memory then a locking semantic is easier to audit. For example, using something like:

```
std::expected<std::unique_ptr<IDataObject>, Error> record = obj.lockNextRecord();
```
provides the user with clear semantics that the caller has exclusive access to the data object, if it is returned. Errors could include "already locked" which would prevent concurrent, mutable access to the data object. To allow multiple readers something like this could be provided:

```
std::expected<std::unique_ptr<const IDataObject>, Error> record = obj.peekNextRecord();
```

The data objects themselves should be created and deleted at well-defined times in the program. This prevents random fluctuations in WCET based on what order shared pointers were acquired and released and by which threads. Instead, given the run-to-completion model libcyphal will enable, we should define when objects will be deleted and check for outstanding locks on this memory as part of this phase. For example: data object deletion could be performed as part of unwinding an IRunnable callstack and any data objects still locked could result in errors. This means any memory leaks would be found within the context of a single run cycle.

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+class ProtocolParameters {
+    +transfer_id_modulo: u64
+    +max_nodes: u16
+    +mtu: u32
+}
+
+%% RUNNABLE
+class IRunnable {
+    <<interface>>
+    +run(now: TimePoint)
```
thirtytwobits, line 284: Can we use chrono please?

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+
+The `DynamicBuffer` provides a uniform API for dealing with the Cyphal transfer payload returned by a lizard and also implements the movable/non-copyable RAII semantics for freeing the memory allocated for the buffer once the dynamic buffer instance is disposed of. The interface hides the gather-scatter nature of the buffer, providing a simplified linearized view. The definition of the class is approximately as follows:
+
+```c++
+/// The buffer is movable but not copyable because copying the contents of a buffer is considered wasteful.
+/// The buffer behaves as if it's empty if the underlying implementation is moved away.
+class DynamicBuffer final
+{
+public:
+    static constexpr std::size_t ImplementationFootprint = sizeof(void*) * 8;
```
thirtytwobits, line 608: I feel like we're not using PMR and CETL enough such that we'll have multiple solutions to the same problem. Take a look at the following examples for how you can manage buffers using unique_ptr and PMR:

1. https://github.com/OpenCyphal/CETL/blob/main/cetlvast/suites/docs/examples/example_07_polymorphic_alloc_deleter.cpp
2. https://github.com/OpenCyphal/CETL/blob/main/cetlvast/suites/docs/examples/example_06_memory_resource_deleter.cpp

And here's how to wrap a C allocator in `memory_resource`: https://github.com/OpenCyphal/CETL/blob/efc3904c5a73550c32809051cf78e1b428271ee8/include/cetl/pmr/o1heap_memory_resource_delegate.hpp#L32

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+    /// The payload may be fragmented to minimize data copying in the user space, allowing the implementation to
+    /// use vectorized I/O (iov).
+    virtual [[nodiscard]] std::expected<bool, std::variant<PlatformError, ArgumentError>>
+    send(const UDPIPEndpoint&                              multicast_endpoint,
+         const std::uint8_t                                dscp,
+         const std::span<const std::span<const std::byte>> payload_fragments) = 0;
+
+    /// This is the Cyphal-layer MTU, which is the maximum number of payload bytes per Cyphal frame.
+    /// To guarantee a single frame transfer, the maximum payload size shall be 4 bytes less to accommodate the CRC.
+    /// The MTU is set per Tx socket individually and it must remain constant after initialization.
```
thirtytwobits, line 752: how will you handle changes to the system while running?

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+    Value         value;
+    RegisterFlags flags;
+};
+
+enum class SetError : std::uint8_t
+{
+    Existence,   ///< Register does not exist.
+    Mutability,  ///< Register is immutable.
+    Coercion,    ///< Value cannot be coerced to the register type.
+    Semantics,   ///< Rejected by the register semantics (e.g. out of range, inappropriate value, bad state, etc).
```
thirtytwobits, line 1247: do we need `Extortion` or `Intimidation` in this list?

docs/design/readme.md
thirtytwobits: isn't it the counterpart of `damn()`?

docs/design/readme.md
pavel-kirienko: cv-qualifiers don't survive copying, so returning `const T` is equivalent to returning `T`. Here, `T = const void*`.

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+
+Type erasure is used throughout to reduce coupling and uphold encapsulation. In some cases, heap allocation that is often associated with type erasure is avoided with the help of small-object-optimized containers that provide an internal arena to store the type-erased object within. More on this below.
+
+Unconventionally, dynamic typing is leveraged with care in certain platform-specific contexts to uphold encapsulation.
+
+### RTTI and dynamic linking considerations
+
+As will be seen later, certain design features are dependent on the ability to detect identical types at runtime, yet the library must be usable without the RTTI capability of the C++ runtime. The following construct is used to provide a small subset of RTTI features that happens to be sufficient for the needs of the library:
+
+```c++
```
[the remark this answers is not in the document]
pavel-kirienko: >The benefit of this approach is that the GUIDs are all explicit identifiers visible in source and they do not change based on toolchain options

I can see that. A disadvantage is that it requires manual assignment of (pseudo) random type identifiers; one specific risk here (and a very real one, I suppose, considering how pervasive copy-paste errors are), which follows directly from the hands-on nature of this approach, is that a developer may forget to roll dice when defining a new type based on an existing one. In the other approach, the dice will be rolled by the linker.

Why can't we use uint64 for the type ID? If you have a hundred types, the collision probability assuming randomly chosen identifiers is negligible -- 2.68e-16.

Just for the sake of completeness, I should add that, assuming a reproducible build environment, one could derive the type ID automatically at compile time as a hash of the file name and the line number, although the disadvantages are clear.

Usage of a reserved value for `NoType` breaks the orthogonality of the type system; it should be an empty option instead.

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+}
+
+%% RUNNABLE
+class IRunnable {
+    <<interface>>
+    +run(now: TimePoint)
+}
+
+%% DYNAMIC BUFFER
+class DynamicBuffer {
```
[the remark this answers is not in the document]
pavel-kirienko: No, because:

1. span is non-owning;
2. span is contiguous.

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+
+Article [Inheritance Without Pointers](https://www.fluentcpp.com/2021/01/29/inheritance-without-pointers/) provides an interesting approach to implementing type erasure without heap based on `std::any`. LibCyphal makes use of its own heapless implementation of `any` to the same end. The resulting container type is defined as follows:
+
+```c++
+/// A generalized implementation of https://www.fluentcpp.com/2021/01/29/inheritance-without-pointers/
+/// that works with any any.
+/// The instance is always initialized with a valid value, but it may turn valueless if the value is moved.
+/// The Any type can be either std::any or a custom alternative.
+template <typename Iface, typename Any>
+class ImplementationCell final
```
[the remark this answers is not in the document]
pavel-kirienko: Neat. I presume in this case, `IPolymorphicType::queryType` works for any base class of the type that the any is populated with, right?

The disadvantage of the tight integration is that it is not compatible with `std::any`, whose `std::any_cast` is not aware of runtime polymorphism. Leaving `ImplementationCell` as proposed makes the library more composable with non-CETL code.

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+
+### General principles
+
+The presentation layer is the first layer that deals with DSDL-generated types. A conventional design would heavily rely on metaprogramming here, providing generic classes parameterized with a DSDL type. This is not a viable choice for LibCyphal because the minimization of template metaprogramming is one of its core design goals. Hence, the design proposed here intentionally avoids static polymorphism where reasonable, preferring runtime polymorphism instead. To this end, the Nunavut C++ generator needs to be modified to ship an interface called `IComposite` as part of its support library, which is to be implemented by generated types:
+
+```c++
+class IComposite
+{
+public:
+    [[nodiscard]] virtual SerializeResult _serialize_(nunavut::support::bitspan out_buffer) const = 0;
```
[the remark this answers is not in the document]
pavel-kirienko: https://github.com/OpenCyphal/CETL/issues/34

docs/design/readme.md
pavel-kirienko: Publishers operating on the same port share the transfer-ID counter and the transport session. Subscribers operating on the same port share the transport session; additionally, each received message is cloned (in the future cloning can be replaced with some form of sharing to reduce data copying) between the user-visible subscription objects. Clients operating on the same service and same server node share the transfer-ID counter, the transport sessions, and the registry of pending requests/responses.

The user code is not responsible for managing the lifecycle of the hidden shared implementation object because that would introduce coupling between different parts of the user's application, which may be operating on the same port (e.g., multiple components may need to receive data from the same subject or invoke the same service). The lifetime of the hidden object begins when the first object on a port is created and ends when the last object on said port is deleted. Thread safety is a valid concern, but it is not explicitly covered in this document, which is my omission.

docs/design/readme.md
pavel-kirienko: This is chrono.

docs/design/readme.md
pavel-kirienko: This class is not responsible for managing the buffer memory but rather for hiding the specifics of its management from the user. The memory management is implemented by the lizards.

docs/design/readme.md
pavel-kirienko: This design assumes that the MTU of the socket does not change while it's open. We could remove this assumption and query the MTU before every transmission through the socket.

docs/design/readme.md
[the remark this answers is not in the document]
thirtytwobits: > requires manual assignment of (pseudo) random type identifiers

The science behind GUID generation is pretty tight and the probably of collision is considered effectively zero. Copy and paste errors are pernicious when using GUIDs. It is the largest issue when using this style of non-rtti programming. Overuse is also a large issue (see https://en.wikipedia.org/wiki/XPCOM) so the intent here, to allow type erasure of DSDL types and to enable the media layer interfaces to be abstract but introspectable, should be strictly adhered to. Perhaps this means we want to use both techniques? The COM-like approach for any externalize-able objects and the compile-time approach for internals?

> Why can't we use uint64 for the type ID?

Because of the "G" in GUID. (or "U" in UUID). See https://en.wikipedia.org/wiki/Universally_unique_identifier. GUIDs allow distribution of shared libraries in binary form whereas compile-time solutions require complete program recompilation. While this isn't a primary use-case the compile-time option specifically prohibits it. Furthermore, GUIDs can be logged and the log values can be found in a source search instead of requiring binary forensics.

> Usage of a reserved value for NoType breaks the orthogonality of the type system; it should be an empty option instead.

Perhaps I meant VoidType and would rename this.

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
+# LibCyphal design overview
+
+LibCyphal is a high-level header-only wrapper over the existing Cyphal implementation libraries, such as LibCANard and LibUDPard (collectively referred to as lizards :lizard:), implemented in C++14 following high-integrity software development practices. It aims to be a full-featured implementation that comes with batteries included and is easy to integrate into any real-time application, from deeply embedded baremetal MCUs up to a conventional POSIX OS (such as GNU/Linux). However, targets with strict memory constraints (such as low-end MCUs with less than ~100k of RAM) are intentionally excluded from the list of intended applications; for those, one should pick a lizard and use it directly.
```
lydia-at-amazon, line 3: I like lizard 🙂 But I also think libard is a liitle more descriptive. Not really a big deal either way

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+    // The type is movable and non-copyable. Moving the instance causes the reference stored in SubscriberImpl to be updated.
+    // Alternatively, the type can be made non-movable to reduce complexity.
+    SubscriberBase(const SubscriberBase&)            = delete;
+    SubscriberBase(SubscriberBase&&)                 = default;
+    SubscriberBase& operator=(const SubscriberBase&) = delete;
+    SubscriberBase& operator=(SubscriberBase&&)      = default;
+    virtual ~SubscriberBase() noexcept               = default;
+
+protected:
+    explicit SubscriberBase(const SharedPtr<PublisherImpl>& impl);
```
lydia-at-amazon, line 1036: Should this be  `SharedPtr<SubscriberImpl>` instead of  `SharedPtr<PublisherImpl>`?

docs/design/readme.md
```diff
@@ -0,0 +1,1704 @@
...
+```c++
+/// The footprint specifies the maximum size of the callable in bytes.
+/// An attempt to use a larger callable will result in a compile-time error.
+template <std::size_t Footprint, typename T, typename... Args>
+class Function final;
+```
+
+### PMR-enabled shared pointer
+
+Certain parts of the presentation layer, possibly along with other parts of the library, will share certain heap-allocated resources internally. While it is possible to implement reference counting ad-hoc, it is desirable to extract this concern into a separate entity, said entity being a PMR-aware shared pointer implementation. By "PMR-aware," we mean that the pointer takes a reference to `std::pmr::memory_resource` upon construction and uses that memory resource to allocate the reference counter instead of using the free store. Similarly, a `makeShared` factory is needed that takes a PMR reference in a similar fashion to allocate the reference counter in the same memory block with the shared object.
```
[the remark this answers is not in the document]
thirtytwobits: Sure. Let's plan on co-developing this. We can work on a branch of CETL until the new features we add there have been stabilized for general use.

docs/design/readme.md
thirtytwobits: This is a more general problem we'll need to plan for. On non-embedded platforms interfaces can come and go, can change their configuration, and generally be chaotic. Trying to handle every possible change specifically is probably not tenable. I imagine we could have some sort of "reconfigure on change" behaviour where, for any given change that affects established parameters (e.g. MTU) we simply repeat the initialization logic used to acquire the media resource in the first place.

lydia-at-amazon:
> TODO:
> 
>     * [ ]  Add non-typed versions of `Publisher`/`Subscriber`/`Client`/`Server` and corresponding factory methods to the `Presentation` controller to allow type-agnostic usage of the presentation layer.

Thanks for adding this!

docs/design/readme.md
pavel-kirienko: The redundant interface class is helpful for runtime reconfiguration

docs/design/readme.md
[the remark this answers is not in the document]
pavel-kirienko: Moving to https://github.com/OpenCyphal/CETL/issues/51

docs/design/readme.md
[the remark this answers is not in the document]
pavel-kirienko: Moving to https://github.com/OpenCyphal/CETL/issues/82

docs/design/readme.md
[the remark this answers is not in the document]
pavel-kirienko: Per the call, I would like to keep this helper. It is not expected to be exposed at the API.

pavel-kirienko:
I think I have addressed everything that we discussed here and at the call last Friday. Unless there are major changes requested, we should be able to start working on the basic components in CETL.

docs/design/readme.md
```diff
@@ -0,0 +1,1802 @@
...
+    /// This is used by Runner implementations to extract the specific underlying platform resource
+    /// (e.g., a file descriptor) from the instance with the help of strict_dynamic_cast<>().
+    virtual TypeID type() const noexcept = 0;
+};
+
+class IPollSet
+{
+public:
+    virtual void addPollable(const IPollable& pollable) = 0;   ///< Block until I/O is possible.
+    virtual void addDeadline(const TimePoint deadline) = 0;  ///< Block untli deadline.
```
lydia-at-amazon, line 496: nit: `untli` --> `until`

lydia-at-amazon (approved):

docs/design/readme.md
```diff
@@ -0,0 +1,1802 @@
...
+/// This is to be done by the implementation, outside of this interface.
+/// All sockets operate only on multicast data.
+/// The sockets produced by this interface shall not outlive it.
+class IMedia
+{
+public:
+    /// LibUDPard requires only one Tx socket per iface.
+    /// We could construct one statically and pass it over during transport initialization,
+    /// but that would be poorly compatible with listen-only nodes that are not interested in sending
+    /// any transfers, as they require no Tx sockets at all.
```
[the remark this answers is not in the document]
pavel-kirienko: Pros: see above

Cons:

1. The user has to manage a transfer-ID counter (and priority) manually per subject.
2. The user has to serialize the message object manually (not applicable for the case of typeless publisher).
3. This design is difficult to implement with transports that must maintain a separate resource per subject. One example is `RedundantTransport`; further, new transports may eventually appear (e.g., there could be a separate TCP socket per subject).

A possible alternative to consider is to add a stateless publication method alongside `makePublisher`, but it doesn't help with item 3 above. Another alternative is to reach to the underlying lizard directly.

docs/design/readme.md
```diff
@@ -0,0 +1,1802 @@
...
+
+```c++
+/// The cetl::IPolymorphicType interface is supposed to provide the _type_() method, as described earlier
+/// (sic! underscores needed to avoid name conflicts), used for runtime type identification.
+class IComposite : public cetl::IPolymorphicType
+{
+public:
+    [[nodiscard]] virtual SerializeResult _serialize_(cetl::bitspan out_buffer) const = 0;
+    [[nodiscard]] virtual SerializeResult _deserialize_(cetl::const_bitspan in_buffer) = 0;
+    [[nodiscard]] virtual std::string_view _name_() const noexcept = 0;
```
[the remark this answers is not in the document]
pavel-kirienko: Yes, this is done by the DSDL type developer:

image

This reminds me that we also need `_name_and_version_`, which I will add now.

docs/design/readme.md
```diff
@@ -0,0 +1,1802 @@
...
+
+    /// Sends the request and returns a promise that will be materialized when the response is successfully received.
+    [[nodiscard]] std::expected<ResponsePromise<DynamicBuffer>, Error> request(
+        const std::span<const std::span<const cetl::byte>> req);
+};
+```
+
+The hidden `ClientImpl` instance is created per pair of (service-ID, server node-ID), and it owns an instance of `IRequestTxSession` and `IResponseRxSession` for sending and receiving RPC requests and responses, respectively. Instances of `ClientImpl` store an AVL tree of all instances of `ResponsePromiseBase`; these types are made movable such that the tree is updated transparently to the user when its items are moved. Calling `request` causes the impl object to send the request; upon successful transmission, a new pending response entry is registered. The `IRunnable::run` interface polls the underlying Rx session for transfers; upon successful reception of a transfer, it locates the matching entry in the `ResponsePromiseBase` tree, and if one is found, the transfer is moved there. Once a promise is materialized, it will be waiting for the user to check it.
```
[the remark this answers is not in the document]
pavel-kirienko: Added `ResponsePromise::setCallback`. It looks a bit haphazard to me but it conveys the idea.

docs/design/readme.md
```diff
@@ -0,0 +1,1802 @@
...
+
+## Application layer
+
+The application layer is the only part of LibCyphal that depends on DSDL-generated types. It is built on top of the presentation layer and contains at least the elements listed below. The application layer is less strict in its avoidance of template metaprogramming, assuming that applications where the use of templates is strongly unadvisable can choose to avoid the reliance on this module.
+
+The main part is the `Node` class that wraps all core components of a Cyphal node into a single entity, which includes at least the following essential components:
+
+- A heartbeat publisher.
+- A `uavcan.node.GetInfo` RPC-service implementation.
+- A registry, as described below.
```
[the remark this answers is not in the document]
pavel-kirienko: Per the current design, a registry will be instantiated automatically with the node. A bare-bones application may choose to use the presentation layer directly without involving the application-layer entities.

docs/design/readme.md
```diff
@@ -0,0 +1,1802 @@
...
+                 IMultiplexer&                      mux,  // only if the second execution strategy proposal is chosen
+                 const std::span<IMedia&>           media,
+                 const std::optional<std::uint16_t> local_node_id);
+```
+
+The CAN transport supports homogeneous redundancy; it can be combined with heterogeneous redundancy with the help of `libcyphal::transport::redundant`.
+
+### Cyphal/UDP
+
+Just like Cyphal/CAN, Cyphal/UDP also supports inner homogeneous redundancy, and it can be combined with `libcyphal::transport::redundant` for heterogeneous redundancy. Unlike Cyphal/CAN, Cyphal/UDP needs dynamic instantiation of sockets depending on the requested Cyphal port configuration, so a socket factory interface is defined:
```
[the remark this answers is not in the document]
pavel-kirienko: Added a new section where this is explained.

lydia-at-amazon (approved):
Looks good, thank you!

PR #335, opened 2024-04-03
Make CI (unit tests and docs) green

serges147:


CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ include(cmake/CMakeCommon.cmake REQUIRED)
 project(libcyphal
         VERSION ${LIBCYPHAL_VERSION}
         LANGUAGES CXX C
-        HOMEPAGE_URL https://github.com/OpenCyphal/libcyphal
+        HOMEPAGE_URL https://github.com/OpenCyphal-Garage/libcyphal
```
pavel-kirienko, line 13: we better not forget to update this when the project is ready to graduate from the garage

.github/workflows/tests.yml
```diff
@@ -2,14 +2,22 @@ name: "libcyphal_test"
...
+      - 'issue/*'
+      # todo: temp - remove this later!
+      - 'sshirokov/CI_*'
   pull_request:
-    branches: [ "main" ]
+    branches:
+      - main
+      - 'issue/*'
+      # todo: temp - remove this later!
+      - 'sshirokov/CI_*'
```
pavel-kirienko, line 15: Isn't it about time to remove it now? You can add hashtag-based triggering like in CETL and elsewhere to trigger it at will.

pavel-kirienko (approved):
LGTM except for the lingering temporary build config, see comments

.github/workflows/tests.yml
serges147: I'll do it soon.

PR #337, opened 2024-04-04
First draft of transport interfaces

serges147:
Added:
- `transport::ISession`
- `transport::IMessage[Rx|Ts]Session`
- `transport::I[Request|Response][Rx|Ts]Session`

Also:
- add "std: [14, 17, 20]" to the build matrix
- add hash tag triggering by #verification #docs tags
- strip repo absolute path prefix from doxygen file paths #docs


CONTRIBUTING.md
```diff
@@ -193,6 +193,28 @@ Reviewers, please check the following items when reviewing a pull-request:
...
     * Is the code in the right namespace/class/function?
 
+### Format the sources
+
+Clang-Format may format the sources differently depending on the version used.
+To ensure that the formatting matches the expectations of the CI suite,
+invoke Clang-Format of the correct version from the container (be sure to use the correct image tag):
+
+```
+docker run --rm -v ${PWD}:/repo ghcr.io/opencyphal/toolshed:ts22.4.3 ./build-tools/bin/verify.py build-danger-danger-repo-clang-format-in-place
```
serges147, line 203: This is currently only working if you first go inside docker (`--it`), install nunavut (`pip install nunavut`), and then run `./build-tools/bin/verify.py build-danger-danger-repo-clang-format-in-place` being there. We probably should somehow "teach" the `verify.py` to install this dependency (and after that we can a bit simplify GH actions by removing `pip install nunavut`).

include/libcyphal/transport/transport.hpp
```diff
@@ -0,0 +1,87 @@
...
+    /// plug-and-play node-ID allocation (for example, a CAN transport may disable automatic retransmission).
+    ///
+    /// @return Optional integer representing the local node-ID.
+    ///
+    CETL_NODISCARD virtual cetl::optional<NodeId> getLocalNodeId() const noexcept = 0;
+
+    /// @brief Provides information about the properties of the transport protocol implemented by the instance.
+    ///
+    /// @return `ProtocolParameters` object if the transport is initialized, otherwise ResultCode::UninitializedError.
+    ///
```
pavel-kirienko, line 79:
```suggestion
```

include/libcyphal/transport/transport.hpp
```diff
@@ -0,0 +1,87 @@
...
+///
+class ITransport : public IRunnable
+{
+public:
+    /// @brief Gets (optional) local node-ID assigned to this transport.
+    ///
+    /// The node-ID is set once during initialization of the transport,
+    /// either explicitly (e.g., CAN) or by deriving the node-ID value from the configuration
+    /// of the underlying protocol layers (e.g., UDP/IP).
+    ///
```
pavel-kirienko, line 66: The part about UDP is not correct (it was changed since the PoC implementation that this comment refers to). We _may_ also at one point allow setting the node-ID once after initialization.

```suggestion
```

include/libcyphal/transport/session/svc_sessions.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+};  // ResponseTxParams
+
+/// @brief Declares an abstract Cyphal transport request RX session interface.
+///
+class IRequestRxSession : public ISession
+{
+public:
+    CETL_NODISCARD virtual RequestRxParams getParams() const noexcept             = 0;
+    virtual void                           setTransferIdTimeout(Duration timeout) = 0;
```
pavel-kirienko, line 63: The way I read AUTOSAR, this is needed because immutable entities should be const and declarations should match definitions token-by-token. Maybe I'm misinterpreting something though.

```suggestion
    virtual void                           setTransferIdTimeout(const Duration timeout) = 0;
```

include/libcyphal/transport/session/svc_sessions.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+namespace libcyphal
+{
+namespace transport
+{
+inline namespace session
+{
+
+/// @brief Declares service request RX session parameters.
+///
```
pavel-kirienko, line 22: Again, I think it is better to write no doc comment at all than to write a comment that adds no information.

include/libcyphal/transport/session/svc_sessions.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+};  // IRequestTxSession
+
+/// @brief Declares an abstract Cyphal transport response RX session interface.
+///
+class IResponseRxSession : public ISession
+{
+public:
+    CETL_NODISCARD virtual ResponseRxParams getParams() const noexcept             = 0;
+    virtual void                            setTransferIdTimeout(Duration timeout) = 0;
```
pavel-kirienko, line 82: Maybe this calls for an IRxSession interface, for regularity? We redefine this method in three different interfaces.

include/libcyphal/transport/session/session.hpp
pavel-kirienko, line 1: Are you sure we need a separate namespace for the sessions? I'm worried excessive granularity might make the code harder to read.

pavel-kirienko (changes requested):

docs/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ function (create_docs_target ARG_DOCS_DOXY_ROOT
     set(DOXYGEN_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
     set(DOXYGEN_CONFIG_FILE ${DOXYGEN_OUTPUT_DIRECTORY}/doxygen.config)
     set(DOXYGEN_EXAMPLE_PATH ${ARG_EXAMPLES_PATH})
+    set(DOXYGEN_STRIP_FROM_PATH ${ARG_DOCS_DOXY_ROOT}/../include/libcyphal)
```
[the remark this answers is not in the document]
serges147: It represents how it was exactly #included in the source code. In "session.hpp" I need include "runnable.hpp" as "libcycphal/runnable.hpp" (aka by "absolute" include path) b/c runnable stays above session & transport. At the same time when I need to include lower level headers (like "session.hpp" from "msg_sessions.hpp") I just use "relative" path approach.

include/libcyphal/types.hpp
```diff
@@ -0,0 +1,75 @@
...
+using NodeId = std::uint16_t;
+
+/// @brief `PortId` is a 16-bit unsigned integer that represents a port (subject & service) in a Cyphal network.
+///
+using PortId = std::uint16_t;
+
+/// @brief `TransferId` is a 64-bit unsigned integer that represents a service transfer (request & response)
+/// in a Cyphal network.
+///
+using TransferId = std::uint64_t;
```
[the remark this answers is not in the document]
serges147: ok, I'll move them.

include/libcyphal/transport/transport.hpp
serges147: I'll probably stop copy/pasting these docs - better not have ones than incorrect ones.

pavel-kirienko: exactly :tada:

include/libcyphal/transport/session/svc_sessions.hpp
serges147: AFAIK at declaration these `const` for value copy has no effect on fn prototype - even CLion highlights them as kinda "not needed".

include/libcyphal/transport/session/svc_sessions.hpp
serges147: I need double check what will happen if there are docs for members on an entity but without its own docs...

include/libcyphal/transport/session/svc_sessions.hpp
pavel-kirienko: True, but then we have this rule, of which CLion is not aware:

image

include/libcyphal/transport/session/session.hpp
serges147: it was my kinda natural feeling that sessions are set of "helper" entities of transport. Also, such granularity (in my experience) helps to isolate/distinguish their implementations (without "polluting" transport namespace where its implementation stuff will live...). But I don't have currently strong feeling, so I'm ok to give it up.

serges147:
_>Why the top header in the diagram does not have libcyphal as the first path component?_
I believe whole documentation is all about "libcyphal", so I thought it is redundant to repeat.

include/libcyphal/runnable.hpp
```diff
@@ -0,0 +1,25 @@
...
+namespace libcyphal
+{
+
+class IRunnable
+{
+public:
+    virtual void run(const TimePoint now) = 0;
+
+protected:
+    virtual ~IRunnable() = default;
```
pavel-kirienko, line 20: rule of 5 is not upheld

include/libcyphal/transport/defines.hpp
```diff
@@ -0,0 +1,40 @@
...
+/// in a Cyphal network.
+///
+using TransferId = std::uint64_t;
+
+struct ProtocolParams final
+{
+    NodeId        max_nodes;
+    std::size_t   mtu_bytes;
+    std::uint64_t transfer_id_modulo;
+
```
pavel-kirienko, line 34: not to nitpick but I'm just curious are you adding the blank lines before the closing brace on purpose or is it a happy little accident

pavel-kirienko (approved):

include/libcyphal/transport/defines.hpp
serges147: Actually on purpose, like if I have in the end comment like this "};  // ProtocolParams" then I insert NL (new line) so that this comment and previous line don't "visually collapse".

PR #339, opened 2024-04-08
Issue 336: Declare libcyphal transport interfaces

serges147:
- Added `ITransport::makeMessage[Rx|Tx]Session` methods
- Added `ITransport::make[Request|Response][Rx|Tx]Session` methods
- Added `Priority`, `[Service]TransferMetadata`, `PayloadFragments`, `MessageRxTransfre` and `ServiceRxTransfer` types.
- Added`transport::AnyError`
- Added implementation of `DynamicBuffer`.

Also:
- fixed "Coverage" build flavor, and artifacts upload
- added `libcyphal::UniquePtr` & `libcyphal::Expected`
- fixed "rule of five" deficiency

### DIFF
--- a/include/libcyphal/runnable.hpp
+++ b/include/libcyphal/runnable.hpp
@@ -19,9 +19,9 @@ public:
     virtual void run(const TimePoint now) = 0;
 
 protected:
-    IRunnable() = default;
+    IRunnable()                 = default;
     IRunnable(const IRunnable&) = default;
-    IRunnable(IRunnable&&) = default;
+    IRunnable(IRunnable&&)      = default;
 
     IRunnable& operator=(const IRunnable&) = default;
     IRunnable& operator=(IRunnable&&)      = default;
--- a/include/libcyphal/runnable.hpp
+++ b/include/libcyphal/runnable.hpp
@@ -14,10 +14,17 @@ namespace libcyphal
 class IRunnable
 {
 public:
+    virtual ~IRunnable() = default;
+
     virtual void run(const TimePoint now) = 0;
 
 protected:
-    virtual ~IRunnable() = default;
+    IRunnable() = default;
+    IRunnable(const IRunnable&) = default;
+    IRunnable(IRunnable&&) = default;
+
+    IRunnable& operator=(const IRunnable&) = default;
+    IRunnable& operator=(IRunnable&&)      = default;
 };
 
 }  // namespace libcyphal
--- a/include/libcyphal/transport/defines.hpp
+++ b/include/libcyphal/transport/defines.hpp
@@ -6,6 +6,8 @@
 #ifndef LIBCYPHAL_TRANSPORT_DEFINES_HPP_INCLUDED
 #define LIBCYPHAL_TRANSPORT_DEFINES_HPP_INCLUDED
 
+#include "dynamic_buffer.hpp"
+
 namespace libcyphal
 {
 namespace transport
@@ -68,9 +70,6 @@ struct ServiceTransferMetadata final : TransferMetadata
 // TODO: Maybe have `cetl::byte` polyfill for C++20
 using PayloadFragments = cetl::span<cetl::span<uint8_t>>;
 
-struct DynamicBuffer final
-{};
-
 struct MessageRxTransfer final
 {
     TransferMetadata       metadata;
@@ -80,7 +79,7 @@ struct MessageRxTransfer final
     MessageRxTransfer(TransferMetadata _metadata, cetl::optional<NodeId> _publisher_node_id, DynamicBuffer _payload)
         : metadata{_metadata}
         , publisher_node_id{_publisher_node_id}
-        , payload{_payload}
+        , payload{std::move(_payload)}
     {
     }
 };
@@ -92,7 +91,7 @@ struct ServiceRxTransfer final
 
     ServiceRxTransfer(ServiceTransferMetadata _metadata, DynamicBuffer _payload)
         : metadata{_metadata}
-        , payload{_payload}
+        , payload{std::move(_payload)}
     {
     }
 };
--- a/include/libcyphal/transport/defines.hpp
+++ b/include/libcyphal/transport/defines.hpp
@@ -28,11 +28,74 @@ using TransferId = std::uint64_t;
 
 struct ProtocolParams final
 {
-    std::uint64_t transfer_id_modulo;
-    std::size_t   mtu_bytes;
-    NodeId        max_nodes;
+    TransferId  transfer_id_modulo;
+    std::size_t mtu_bytes;
+    NodeId      max_nodes;
 
-};  // ProtocolParams
+    ProtocolParams(TransferId _transfer_id_modulo, std::size_t _mtu_bytes, NodeId _max_nodes)
+        : transfer_id_modulo{_transfer_id_modulo}
+        , mtu_bytes{_mtu_bytes}
+        , max_nodes{_max_nodes}
+    {
+    }
+};
+
+struct TransferMetadata
+{
+    TransferId transfer_id;
+    TimePoint  timestamp;
+    Priority   priority;
+
+    TransferMetadata(TransferId _transfer_id, TimePoint _timestamp, Priority _priority)
+        : transfer_id{_transfer_id}
+        , timestamp{_timestamp}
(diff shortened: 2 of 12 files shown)


include/libcyphal/runnable.hpp
```diff
@@ -14,10 +14,17 @@ namespace libcyphal
...
     virtual void run(const TimePoint now) = 0;
 
 protected:
-    virtual ~IRunnable() = default;
+    IRunnable()                 = default;
+    IRunnable(const IRunnable&) = default;
+    IRunnable(IRunnable&&)      = default;
+
+    IRunnable& operator=(const IRunnable&) = default;
+    IRunnable& operator=(IRunnable&&)      = default;
```
pavel-kirienko, line 27: I think runnables must be non-copyable, otherwise things like `AggregateRunnable` et all will break, because we have to be able to keep valid references to the interface. If the implementation object is moved, references lose validity.

include/libcyphal/types.hpp
```diff
@@ -16,6 +18,19 @@
 namespace libcyphal
 {
 
+enum class Priority
```
pavel-kirienko, line 21: Transfer priority is also a transport-layer entity, so it should probably go there.

include/libcyphal/transport/transport.hpp
```diff
@@ -19,6 +20,18 @@ class ITransport : public IRunnable
...
+    CETL_NODISCARD virtual Expected<UniquePtr<session::IMessageTxSession>, AnyError> makeMessageTxSession(
+        session::MessageTxParams params) = 0;
+    CETL_NODISCARD virtual Expected<UniquePtr<session::IRequestRxSession>, AnyError> makeRequestRxSession(
+        session::RequestRxParams params) = 0;
+    CETL_NODISCARD virtual Expected<UniquePtr<session::IRequestTxSession>, AnyError> makeRequestTxSession(
+        session::RequestTxParams params) = 0;
+    CETL_NODISCARD virtual Expected<UniquePtr<session::IResponseRxSession>, AnyError> makeResponseRxSession(
+        session::ResponseRxParams params) = 0;
+    CETL_NODISCARD virtual Expected<UniquePtr<session::IResponseTxSession>, AnyError> makeResponseTxSession(
+        session::ResponseTxParams params) = 0;
```
pavel-kirienko, line 34: Shouldn't the parameters be const?

include/libcyphal/transport/session/msg_sessions.hpp
```diff
@@ -29,13 +29,16 @@ struct MessageTxParams final
 class IMessageRxSession : public IRxSession
 {
 public:
-    CETL_NODISCARD virtual MessageRxParams getParams() const noexcept = 0;
+    CETL_NODISCARD virtual MessageRxParams   getParams() const noexcept = 0;
+    CETL_NODISCARD virtual MessageRxTransfer receive()                  = 0;
```
pavel-kirienko, line 33: There's a missing detail in the class diagram although it is deducible from the following description: all `receive` methods should return an optional. Here it would be `cetl::optional<MessageRxTransfer>`. All methods are non-blocking; if there's nothing to return, an empty option is used.

include/libcyphal/transport/defines.hpp
```diff
@@ -28,11 +30,71 @@ using TransferId = std::uint64_t;
...
 struct ProtocolParams final
 {
-    NodeId        max_nodes;
-    std::size_t   mtu_bytes;
-    std::uint64_t transfer_id_modulo;
+    TransferId  transfer_id_modulo;
+    std::size_t mtu_bytes;
+    NodeId      max_nodes;
+
+    ProtocolParams(TransferId _transfer_id_modulo, std::size_t _mtu_bytes, NodeId _max_nodes)
```
pavel-kirienko, line 37: Why do we need an explicit constructor here? Can we not use aggregate initialization? Same question applies to the other definitions.

include/libcyphal/transport/defines.hpp
```diff
@@ -28,11 +30,71 @@ using TransferId = std::uint64_t;
...
+
+    ServiceTransferMetadata(TransferId _transfer_id, TimePoint _timestamp, Priority _priority, NodeId _remote_node_id)
+        : TransferMetadata{_transfer_id, _timestamp, _priority}
+        , remote_node_id{_remote_node_id}
+    {
+    }
+};
+
+// TODO: Maybe have `cetl::byte` polyfill for C++20
+using PayloadFragments = cetl::span<cetl::span<uint8_t>>;
```
pavel-kirienko, line 71: The C++ standard does not define `::uint8_t`

```suggestion
using PayloadFragments = cetl::span<cetl::span<std::uint8_t>>;
```

include/libcyphal/transport/defines.hpp
```diff
@@ -28,11 +30,71 @@ using TransferId = std::uint64_t;
...
+
+struct MessageRxTransfer final
+{
+    TransferMetadata       metadata;
+    cetl::optional<NodeId> publisher_node_id;
```
pavel-kirienko, line 76: I'm wondering if these should have been extracted into a `MessageTransferMetadata` in my design proposal.

test/unittest/transport/test_dynamic_buffer.cpp
```diff
@@ -0,0 +1,124 @@
...
+class InterfaceMock : public DynamicBuffer::Interface
+{
+public:
+    MOCK_METHOD(void, moved, ());
+    MOCK_METHOD(void, deinit, ());
+
+    MOCK_METHOD(std::size_t, size, (), (const override));
+    MOCK_METHOD(std::size_t, copy, (const std::size_t, void* const, const std::size_t), (const override));
+};
+class InterfaceWrapper : public rtti_helper<type_id_type<0x01>, DynamicBuffer::Interface>
```
pavel-kirienko, line 30: Can we use randomly chosen numbers for type IDs please to avoid the risk of collisions when the test suite grows larger? If you chose 0x01 randomly then please ignore this comment.

https://xkcd.com/221

```suggestion
class InterfaceWrapper : public rtti_helper<type_id_type<0x56, 0x85>, DynamicBuffer::Interface>
```

test/unittest/transport/test_dynamic_buffer.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+        if (mock_ != nullptr)
+        {
+            mock_->moved();
+        }
+    }
+
+};  // InterfaceWrapper
+
+TEST(test_dynamic_buffer, move_ctor_assign__size)
```
pavel-kirienko, line 84: This might be a nitpick considering that this is happening in a test suite but identifiers containing double underscore are reserved in C and C++.

pavel-kirienko (changes requested):

include/libcyphal/transport/defines.hpp
serges147: that was typo

include/libcyphal/transport/defines.hpp
serges147: I'll take a closer look...

include/libcyphal/transport/defines.hpp
serges147: Probably we don't need them. It's my swift language habits ;) - we remove them all.

include/libcyphal/transport/defines.hpp
serges147: yup, makes sense.

include/libcyphal/transport/defines.hpp
```diff
@@ -26,13 +28,57 @@ using PortId = std::uint16_t;
...
+};
+
+struct ServiceTransferMetadata final : TransferMetadata
+{
+    NodeId remote_node_id;
+};
+
+// TODO: Maybe have `cetl::byte` polyfill for C++20
+using PayloadFragments = cetl::span<cetl::span<std::uint8_t>>;
```
pavel-kirienko, line 69: Missing const qualifier.

pavel-kirienko (approved):
https://wasteaguid.info/

include/libcyphal/transport/defines.hpp
serges147: will double check (next pr)

PR #340, opened 2024-04-10
Sshirokov/336 transport

serges147:
- Added `ITransport`, `ICanTransport` & `IUpdTransport`
- `PayloadFragment` now immutable; `FragmentBuffer` is mutable span.


include/libcyphal/transport/msg_sessions.hpp
```diff
@@ -49,14 +47,13 @@ class IMessageTxSession : public IRunnable
     ///
     /// @param metadata Additional metadata associated with the message.
     /// @param payload_fragments Segments of the message payload.
-    /// @return `void` in case of success; otherwise an error.
+    /// @return `nullopt` in case of success; otherwise a transport error.
     ///
-    CETL_NODISCARD virtual Expected<void, AnyError> send(const TransferMetadata& metadata,
+    CETL_NODISCARD virtual cetl::optional<AnyError> send(const TransferMetadata& metadata,
```
serges147, line 52: Making `void` specialization for `Expected` in my near future todo list. Temporary just using `optional`.

include/libcyphal/transport/multiplexer.hpp
```diff
@@ -0,0 +1,20 @@
...
+#ifndef LIBCYPHAL_TRANSPORT_MULTIPLEXER_HPP_INCLUDED
+#define LIBCYPHAL_TRANSPORT_MULTIPLEXER_HPP_INCLUDED
+
+namespace libcyphal
+{
+namespace transport
+{
+
+class IMultiplexer
+{};
```
serges147, line 15: Such empty interface is just temporary - will be in use soon.

include/libcyphal/transport/can/transport.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+class ICanTransport : public ITransport
+{};
+
+struct Factory final
+{
+public:
+    Factory() = delete;
+
+    CETL_NODISCARD static inline Expected<UniquePtr<ICanTransport>, FactoryError> make(
```
pavel-kirienko, line 28: Why can't this be a freestanding function?

pavel-kirienko (changes requested):

include/libcyphal/transport/can/media.hpp
```diff
@@ -0,0 +1,87 @@
...
+    /// The lifetime of the filter array may end upon return (no references retained).
+    /// @return Returns `true` on success, `false` in case of a low-level error (e.g., IO error).
+    ///
+    CETL_NODISCARD virtual bool setFilters(const Filters filters) noexcept = 0;
+
+    /// @brief Schedules the frame for transmission asynchronously and return immediately.
+    ///
+    /// @return Returns `true` if accepted or already timed out; `false` to try again later.
+    ///
+    CETL_NODISCARD virtual Expected<bool, cetl::variant<ArgumentError>> push(
```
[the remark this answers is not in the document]
serges147: ok, I'll add it

include/libcyphal/transport/defines.hpp
```diff
@@ -66,7 +66,15 @@ struct ServiceTransferMetadata final : TransferMetadata
...
-using PayloadFragments = cetl::span<cetl::span<std::uint8_t>>;
+
+/// @brief Defines a mutable span of bytes that represents a buffer for payload fragment.
+using FragmentBuffer = cetl::span<std::uint8_t>;
+
+/// @brief Defines a span of bytes that represents an immutable fragment of payload.
+using PayloadFragment = cetl::span<const std::uint8_t>;
+
+/// @brief Defines a span of immutable fragments of payload.
+using PayloadFragments = cetl::span<const PayloadFragment>;
```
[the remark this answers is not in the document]
serges147: ok, I will give up aliases for trivial `cetl::span<[const] std::uint8_t>` (aka `FragmentBuffer` & `PayloadFragment `), but keep `PayloadFragments`.

include/libcyphal/transport/udp/transport.hpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+    CETL_NODISCARD ProtocolParams getProtocolParams() const noexcept override
+    {
+        return ProtocolParams{};
+    }
+
+    CETL_NODISCARD Expected<UniquePtr<IMessageRxSession>, AnyError> makeMessageRxSession(
+        const MessageRxParams&) override
+    {
+        return ArgumentError{};
```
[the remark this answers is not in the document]
serges147: yes, this is the idea, I just not yet looked at UDP at all - anyway, I will replace them.

include/libcyphal/transport/can/transport.hpp
serges147: My rationale was to avoid freestanding functions (which for my taste a bit "c-style"), but instead you know "encapsulate" them into some sort of entity (like this `Factory`). Also, IMHO if we apply such pattern everywhere then it will be easier for users to guess where to find a factory method, like always `<correct_namespace>::Factory::make(...)` something. But in general I'm totally good with essentially the same `<correct_namespace>::makeSomething(...)` approach.

Probably it's again my Swift habits - there are no namespaces in Swift, so any freestanding function is GLOBAL one, which is considered as bad design - so you nest such functions in some (sub)type as static one.

Please let me know if my rationale is not applicable for c++, or there are reasons to specifically go with freestanding alternative. Thanks!

pavel-kirienko (approved):

include/libcyphal/transport/can/transport.hpp
serges147: will fix this in the next pr

PR #341, opened 2024-04-11
First integration with canard.

serges147:
- `canardInit` is in use.
- CAN transport memory resource is connected to the canard allocate/free.
- Implemented `getLocalNodeId`

Also:
- Get rid of extra `Factory` entities.


include/libcyphal/transport/can/transport.hpp
```diff
@@ -20,36 +22,34 @@ namespace can
...
+    }
+
     // ICanTransport
 
     // ITransport
 
     CETL_NODISCARD cetl::optional<NodeId> getLocalNodeId() const noexcept override
     {
-        return cetl::nullopt;
+        return canard_instance_.node_id == CANARD_NODE_ID_UNSET ? cetl::nullopt
```
pavel-kirienko, line 46:
```suggestion
        return canard_instance_.node_id > CANARD_NODE_ID_MAX ? cetl::nullopt
```

include/libcyphal/transport/can/transport.hpp
```diff
@@ -20,36 +22,34 @@ namespace can
...
+                                                                : cetl::make_optional(canard_instance_.node_id);
     }
     CETL_NODISCARD ProtocolParams getProtocolParams() const noexcept override
     {
-        return ProtocolParams{};
+        // TODO: Fill MTU
+        return ProtocolParams{1 << CANARD_TRANSFER_ID_BIT_LENGTH, 0, CANARD_NODE_ID_MAX + 1};
```
pavel-kirienko, line 52: I would use designated initializers here: `.mtu=...` etc

include/libcyphal/transport/can/transport.hpp
```diff
@@ -87,11 +87,65 @@ class TransportImpl final : public ICanTransport
...
+    // see https://github.com/OpenCyphal/libcanard/issues/216
+    //
+    struct CanardMemory
+    {
+        std::size_t size;
+        alignas(std::max_align_t) cetl::byte data[1];
+    };
+
+    cetl::pmr::memory_resource& memory_;
+    CanardInstance              canard_instance_;
```
pavel-kirienko, line 103: Shouldn't the fields go to the end of the class scope?

include/libcyphal/transport/can/transport.hpp
```diff
@@ -20,36 +22,34 @@ namespace can
...
 namespace detail
 {
 
 class TransportImpl final : public ICanTransport
 {
 public:
+    TransportImpl(cetl::pmr::memory_resource& memory, const CanardNodeID canard_node_id)
+        : memory_{memory}
+    {
+        canard_instance_ = canardInit(canardMemoryAllocate, canardMemoryFree);
```
pavel-kirienko, line 34:
```suggestion
        : memory_{memory}
        , canard_instance_(canardInit(canardMemoryAllocate, canardMemoryFree))
    {
```

include/libcyphal/transport/can/transport.hpp
```diff
@@ -87,11 +87,65 @@ class TransportImpl final : public ICanTransport
...
+        CETL_DEBUG_ASSERT(ins->user_reference != nullptr, "Expected `this` transport as user reference.");
+
+        return *static_cast<TransportImpl*>(ins->user_reference);
+    }
+
+    CETL_NODISCARD static void* canardMemoryAllocate(CanardInstance* ins, size_t amount)
+    {
+        auto& self = getSelfFrom(ins);
+
+        const auto memory_size   = offsetof(CanardMemory, data) + amount;
```
pavel-kirienko, line 117: AUTOSAR M18-2-1, another occurrence below

pavel-kirienko (changes requested):

include/libcyphal/transport/can/transport.hpp
pavel-kirienko: Ah yes sorry

include/libcyphal/transport/can/transport.hpp
serges147: they are in private section, so I though it is fine place. I'm I right that you would like almost always see PRIVATE stuff in the following order:
1. private types, aliases
2. private member functions
3. private member consts and vars
?
maybe it worth to document this at https://kb.zubax.com/pages/viewpage.action?pageId=2195699 ?

include/libcyphal/transport/can/transport.hpp
```diff
@@ -87,11 +86,69 @@ class TransportImpl final : public ICanTransport
...
+    static void canardMemoryFree(CanardInstance* ins, void* pointer)
+    {
+        if (pointer == nullptr)
+        {
+            return;
+        }
+
+        const auto uint_ptr = reinterpret_cast<std::uintptr_t>(pointer);
+        CETL_DEBUG_ASSERT(uint_ptr > CanardMemory::DataOffset, "Invalid too small pointer.");
+        const auto canard_memory = reinterpret_cast<CanardMemory::Layout*>(uint_ptr - CanardMemory::DataOffset);
```
pavel-kirienko, line 138: The conversion to integer and the DataOffset all look a bit shady but we will accept this for now pending proper solution. I think it would have been a lot easier to just go like:

```c++
struct CanardMemory
{
    alignas(std::max_align_t) std::size_t size;
};
static_assert(sizeof(CanardMemory) == alignof(std::max_align_t));
```

And then:

```c++
auto* const mem = static_cast<CanardMemory*>(pointer)-1;
```

But it's good enough as a tentative solution.

pavel-kirienko (approved):

PR #342, opened 2024-04-12
Implemented `transport::can::TransportImpl::getProtocolParams`

serges147:
- `cetl::VariableLengthArray` is now in use to store CAN transport media interfaces.
- min MTU from all media is reported by the `ICanTransport::getProtocolParams`.
- added corresponding unit tests (to cover `getProtocolParams`).

Also:
- a bit reworked (simplified) canard memory alloc/free

### DIFF
--- a/include/libcyphal/transport/can/transport.hpp
+++ b/include/libcyphal/transport/can/transport.hpp
@@ -26,18 +26,12 @@ class ICanTransport : public ITransport
 
 namespace detail
 {
-template <typename T>
-using VarArray = libcyphal::detail::VarArray<T>;
-
-template <typename T>
-using PmrAllocator = libcyphal::detail::PmrAllocator<T>;
-
 class TransportImpl final : public ICanTransport
 {
 public:
-    TransportImpl(cetl::pmr::memory_resource& memory,
-                  VarArray<IMedia*>&&         media_array,
-                  const CanardNodeID          canard_node_id)
+    TransportImpl(cetl::pmr::memory_resource&            memory,
+                  libcyphal::detail::VarArray<IMedia*>&& media_array,
+                  const CanardNodeID                     canard_node_id)
         : memory_{memory}
         , media_array_{std::move(media_array)}
         , canard_instance_{canardInit(canardMemoryAllocate, canardMemoryFree)}
@@ -173,9 +167,9 @@ private:
 
     // MARK: Data members:
 
-    cetl::pmr::memory_resource& memory_;
-    const VarArray<IMedia*>     media_array_;
-    CanardInstance              canard_instance_;
+    cetl::pmr::memory_resource&                memory_;
+    const libcyphal::detail::VarArray<IMedia*> media_array_;
+    CanardInstance                             canard_instance_;
 
 };  // TransportImpl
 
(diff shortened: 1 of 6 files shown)


include/libcyphal/transport/can/transport.hpp
```diff
@@ -24,21 +26,29 @@ class ICanTransport : public ITransport
...
         : memory_{memory}
+        , media_array_{std::move(media_array)}
         , canard_instance_{canardInit(canardMemoryAllocate, canardMemoryFree)}
     {
         canard_instance_.user_reference = this;
         canard_instance_.node_id        = canard_node_id;
     }
 
-    // ICanTransport
+    // MARK: ICanTransport
```
serges147, line 49: CLion nicely detects these `// MARK: xxx` comments:
Screenshot 2024-04-12 at 17 24 53

include/libcyphal/transport/can/transport.hpp
```diff
@@ -133,40 +143,75 @@ class TransportImpl final : public ICanTransport
...
 
+    template <typename Action>
+    void forEachMedia(Action action)
+    {
+        for (const auto media : media_array_)
+        {
+            CETL_DEBUG_ASSERT(media != nullptr, "Expected media interface.");
+            action(std::ref(*media));
+        }
+    }
```
serges147, line 161: not in use yet but I anticipate the need

include/libcyphal/transport/can/transport.hpp
```diff
@@ -133,40 +143,75 @@ class TransportImpl final : public ICanTransport
...
+
+    template <typename T, typename Reducer>
+    T reduceMediaInto(T&& init, Reducer reducer) const
+    {
+        for (const auto media : media_array_)
+        {
+            CETL_DEBUG_ASSERT(media != nullptr, "Expected media interface.");
+            reducer(std::forward<T>(init), std::ref(*media));
```
pavel-kirienko, line 169: It is an interesting choice that the reducer modifies the argument in-place. Conventionally (including std::accumulate and std::reduce) the function returns the new value. std::for_each is different but its main purpose is to invoke side effects so it makes sense there.

include/libcyphal/transport/can/transport.hpp
```diff
@@ -133,40 +143,75 @@ class TransportImpl final : public ICanTransport
...
     if (local_node_id.has_value() && local_node_id.value() > CANARD_NODE_ID_MAX)
     {
         return ArgumentError{};
     }
-    const auto canard_node_id = static_cast<CanardNodeID>(local_node_id.value_or(CANARD_NODE_ID_UNSET));
 
-    cetl::pmr::polymorphic_allocator<detail::TransportImpl> allocator{&memory};
+    detail::VarArray<IMedia*> media_array{MaxMediaInterfaces, &memory};
+    media_array.reserve(static_cast<std::size_t>(media_count));
+    std::copy_if(media.cbegin(), media.cend(), std::back_inserter(media_array), [](auto m) { return m != nullptr; });
```
pavel-kirienko, line 209:
```suggestion
    std::copy_if(media.cbegin(), media.cend(), std::back_inserter(media_array), [](auto m) { return m != nullptr; });
    assert(!media_array.empty() && (media_array.size() == media_count));
```

pavel-kirienko (approved):

include/libcyphal/transport/can/transport.hpp
serges147: There actually exactly two (non-exclusive) approaches:
1. "conventional" reduce - return new value
2. and reduce "into".

You normally you use #1 for simple primitive types; and #2 for more complex/expensive stuff (like collection f.e.).
Look like currently my case is #1, but I wanted to make `reduceMediaInto` a bit more universal. So, if don't mind, I will keep it like it's currently, but will keep eye on this and refactor it (especially if there will be no #2 usages lately). ok?

pavel-kirienko (approved):

PR #343, opened 2024-04-15
Issue/336 transport iterfaces

serges147:
First draft of transport interfaces

Added:
- `transport::ISession`
- `transport::IMessage[Rx|Ts]Session`
- `transport::I[Request|Response][Rx|Ts]Session`

Also:
- add "std: [14, 17, 20]" to the build matrix
- add hash tag triggering by #verification #docs tags
- strip repo absolute path prefix from doxygen file paths


include/libcyphal/transport/defines.hpp
```diff
@@ -0,0 +1,89 @@
...
+};
+
+struct ServiceRxTransfer final
+{
+    ServiceTransferMetadata metadata;
+    DynamicBuffer           payload;
+};
+
+/// @brief Defines maximum number of media interfaces that can be used in a Cyphal transport.
+constexpr std::size_t MaxMediaInterfaces = 3;
```
pavel-kirienko, line 84: We are not going to need this after the cetl::span issue is resolved, right? Maybe it's worth adding a clarifying comment here about this.

test/unittest/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ foreach(NATIVE_TEST ${NATIVE_TESTS})
             ROOT_DIRECTORY ${LIBCYPHAL_ROOT}
             TARGET_EXECUTION_DEPENDS ${LOCAL_TEST_REPORT}
             OBJECT_LIBRARY ${LOCAL_TEST_LIB}
+            EXCLUDE_PATHS ${LIBCYPHAL_ROOT}external
```
pavel-kirienko, line 52: Are we missing an `/` here? I get it that `LIBCYPHAL_ROOT` contains a trailing `/` but it's a matter of readability

pavel-kirienko (dismissed):

pavel-kirienko (dismissed):

include/libcyphal/transport/can/media.hpp
```diff
@@ -0,0 +1,88 @@
...
+using Filters = cetl::span<const Filter>;
+
+struct RxMetadata final
+{
+    TimePoint   timestamp;
+    CanId       can_id;
+    std::size_t payload_size;
+};
+
+class IMedia
```
thirtytwobits, line 36: We're missing a key feature I wanted to have where we implement "borrowed memory" that we ask for from the media layer and use that to write frames into. This enables implementations using DMA buffers where the peripheral TX buffer can be borrowed and used directly by our message serializers thus enabling no-copy serialization of data structures to be transmitted directly by hardware. It's easy enough to simply provide regular memory buffers and copy into sockets when implementing a media layer for something like linux but the borrowed-memory model allows MCUs to have incredibly efficient integrations when needed.  I want this to be a common pattern used by all media layer implementations regardless of transport.

One way to implement this is [DirectX style](https://learn.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dvertexbuffer9-lock) where the act of successfully locking memory provides the caller a pointer they can use until they unlock. This works well for DirectX since the act of unlocking becomes an event the 3d engine uses to release the memory to the GPU.

For libcyphal, I'd rather find a way to implement a "media allocator" that is a polymorphic allocator obtained from the media layer and populated with a `memory_resource` implemented in terms of DMA, DTCM, plain-old-system memory, etc (i.e. the simplest linux media layer would simply use the system's new_delete_resource). The challenge with this is there is no "I'm done with this, please submit to hardware" event that is intrinsic to the mechanism. Overloading the deallocate method to do this seem ill-advised (i.e. "to send your message, delete it" is not really a good plan). We could actually subclass polymorphic_allocator and add lock/unlock region methods to it?

include/libcyphal/transport/can/transport.hpp
```diff
@@ -0,0 +1,219 @@
...
+class ICanTransport : public ITransport
+{};
+
+namespace detail
+{
+class TransportImpl final : public ICanTransport
+{
+public:
+    TransportImpl(cetl::pmr::memory_resource&            memory,
+                  libcyphal::detail::VarArray<IMedia*>&& media_array,
```
thirtytwobits, line 33: why are we aliasing variable_length_array but not span? What's the logic we're using to make a type a libcyphal type rather than an external one?

include/libcyphal/transport/can/transport.hpp
```diff
@@ -0,0 +1,219 @@
...
+                  const CanardNodeID                     canard_node_id)
+        : memory_{memory}
+        , media_array_{std::move(media_array)}
+        , canard_instance_{canardInit(canardMemoryAllocate, canardMemoryFree)}
```
thirtytwobits, line 37: Can canardInit fail?

include/libcyphal/transport/can/transport.hpp
```diff
@@ -0,0 +1,219 @@
...
+    {
+        return NotImplementedError{};
+    }
+    CETL_NODISCARD Expected<UniquePtr<IResponseTxSession>, AnyError> makeResponseTxSession(
+        const ResponseTxParams&) override
+    {
+        return NotImplementedError{};
+    }
+
+    // MARK: IRunnable
```
thirtytwobits, line 90: Who's Mark? Do I know him?

test/unittest/transport/can/media_mock.hpp
```diff
@@ -0,0 +1,40 @@
+/// @copyright
+/// Copyright (C) OpenCyphal Development Team  <opencyphal.org>
+/// Copyright Amazon.com Inc. or its affiliates.
+/// SPDX-License-Identifier: MIT
+
+#ifndef LIBCYPHAL_TRANSPORT_CAN_MEDIA_MOCK_HPP_INCLUDED
```
thirtytwobits, line 6: love seeing these mocks! Thank you.

thirtytwobits (commented):
There's a complete lack of interesting ASCII art in this source so far. Very boring!

include/libcyphal/transport/can/transport.hpp
pavel-kirienko: image

include/libcyphal/transport/can/transport.hpp
pavel-kirienko: image

include/libcyphal/transport/can/transport.hpp
pavel-kirienko: >why are we aliasing variable_length_array but not span?

It's a workaround for https://github.com/OpenCyphal/CETL/issues/115 due to be undone.

@serges147 can we please use span as originally planned, considering that CETL has it fixed now?

include/libcyphal/transport/can/media.hpp
```diff
@@ -0,0 +1,88 @@
...
+    /// @brief Set the filters for the CAN bus.
+    ///
+    /// If there are fewer hardware filters available than requested, the configuration will be coalesced as described
+    /// in the Cyphal/CAN Specification. If zero filters are requested, all incoming traffic will be rejected.
+    /// While reconfiguration is in progress, incoming frames may be lost and/or unwanted frames may be received.
+    /// The lifetime of the filter array may end upon return (no references retained).
+    ///
+    /// @return Returns `true` on success, `false` in case of a low-level error (e.g., IO error).
+    ///
+    CETL_NODISCARD virtual bool setFilters(const Filters filters) noexcept = 0;
```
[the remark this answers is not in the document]
pavel-kirienko: >Why not MediaError here?

Ah yes, good point. @serges147 I think we need an expected here with appropriate error options: PlatformError (I think this is what Scott was referring to instead of MediaError as we don't have that error type), ArgumentError, CapacityError.

>Also, should we add a bool do_not_coalescence argument that means the set will fail if there aren't enough hardware filters when this is true?

No, but we could add `do_not_coalesce` :) If coalescence is disabled and there are not enough filters you get CapacityError.

include/libcyphal/transport/can/media.hpp
```diff
@@ -0,0 +1,88 @@
...
+    CETL_NODISCARD virtual std::size_t getMtu() const noexcept = 0;
+
+    /// @brief Set the filters for the CAN bus.
+    ///
+    /// If there are fewer hardware filters available than requested, the configuration will be coalesced as described
+    /// in the Cyphal/CAN Specification. If zero filters are requested, all incoming traffic will be rejected.
+    /// While reconfiguration is in progress, incoming frames may be lost and/or unwanted frames may be received.
+    /// The lifetime of the filter array may end upon return (no references retained).
+    ///
+    /// @return Returns `true` on success, `false` in case of a low-level error (e.g., IO error).
```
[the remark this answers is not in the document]
pavel-kirienko: This will be addressed with polymorphic PlatformError as discussed in https://github.com/OpenCyphal-Garage/libcyphal/issues/345

include/libcyphal/transport/dynamic_buffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+namespace libcyphal
+{
+namespace transport
+{
+
+/// The buffer is movable but not copyable because copying the contents of a buffer is considered wasteful.
+/// The buffer behaves as if it's empty if the underlying implementation is moved away.
+///
+class DynamicBuffer final
```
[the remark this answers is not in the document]
pavel-kirienko: We're circling back to the same conversation again: we do in fact need this DynamicBuffer because it is designed to provide a contiguous view of fragmented lizard-specific buffers. It is not flat. It erases the type information of the underlying buffer and makes it appear contiguous. Please refer here for details: https://github.com/OpenCyphal-Garage/libcyphal/blob/main/docs/design/readme.md#dynamicbuffer

include/libcyphal/transport/msg_sessions.hpp
```diff
@@ -0,0 +1,59 @@
...
+    /// @brief Receives a message from the transport layer.
+    ///
+    /// Method is not blocking, and will return immediately if no message is available.
+    ///
+    /// @return A message transfer if available; otherwise an empty optional.
+    ///
+    CETL_NODISCARD virtual cetl::optional<MessageRxTransfer> receive() = 0;
+};
+
+class IMessageTxSession : public IRunnable
```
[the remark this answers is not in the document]
pavel-kirienko: This is sensible but:

- Sometimes it is necessary to provide more holistic documentation covering multiple entities at once instead of scattering it per entity. This is one such case. We could copy-paste relevant parts of the design doc to the `ITransport` docs in `transport.hpp` and put references to that in the class comments here. The relevant sections are:
   - https://github.com/OpenCyphal-Garage/libcyphal/blob/main/docs/design/readme.md#session-factory-methods
   - https://github.com/OpenCyphal-Garage/libcyphal/blob/main/docs/design/readme.md#session-data-flow

- Some methods do not require documentation. If you have `Thing getThing() const noexcept`, a doc comment saying "Gets the thing" adds nothing but noise. I would prefer to keep obvious cases undocumented.

include/libcyphal/transport/can/media.hpp
pavel-kirienko: This is not going to be an easy change at the moment because it affects the lizards. The transmission pipelines in both libudpard and libcanard are currently copying. You give the library your transfer payload (which currently has to be a contiguous buffer but it is quite easy to support scattered buffers which yields certain benefits), it allocates a sequence of frame payload buffers and copies fragments of your original buffer into them, adding frame headers/tails in the process. It is possible to eliminate the copying if we could allow the enqueued tx frame items (see `UdpardTxItem`, `CanardTxQueueItem`) to reference memory in the original transfer payload instead of containing own copes of it; said tx queue item objects will still need to be heap-allocated, though, but they would become very small.

To support this, the media layer implementations would have to be able to accept not just contiguous frame payloads (currently we push to the media layer complete buffers ready to be written into the socket or whatever) but fragments of them, so something similar to POSIX iov will be needed.

Most importantly, LibCyphal will need to keep the transfer payload buffer alive until all frames that originated from it have cleared the tx queue (either by being transmitted or by timing out). This is actually not super hard to implement because 1. Lizards provide continuity between the transfer pushed by the user and the frames that originate from it (see the `user_transfer_reference` field); and 2. Lizards return the number of frames enqueued per transfer; so we can implement reference counting quite easily.

Earlier, I mentioned that scattered transfer payload buffers offer "certain benefits" during transmission; what I mean is that if we support scattered buffers, DSDL serialization could be sped up considerably in cases where there are large byte arrays present in the message (like imagery data) because the serializer would be able to serialize only non-trivial parts of the message and keep references to the blobs:

```python
uavcan.time.SynchronizedTimestamp.1.0 timestamp  # NON ZERO-COPY
uint16 IMAGE_WIDTH  = 1280
uint16 IMAGE_HEIGHT = 800
uint8 BITS_PER_PIXEL = 24
uint8[IMAGE_WIDTH * IMAGE_HEIGHT * BITS_PER_PIXEL / 8] pixel_data  # ZERO-COPY!
```

If we go this way, to support full end-to-end low-copy operation, the application would need to allocate the message object itself from the media memory resource.

One side effect of scattered input buffers is that an output frame may end up straddling multiple fragments; to support this we will need to allow a tx queue item to reference not just one but an arbitrary number of fragments from the input buffer.

Am I thinking in the correct direction here?

Note that there are related considerations affecting the RX pipeline:

image

include/libcyphal/transport/defines.hpp
serges147: Please let me apply fix in separate dedicated small PR.

include/libcyphal/transport/msg_sessions.hpp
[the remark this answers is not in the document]
serges147: As, I wrote for other PR, documentation for public stuff is definitely coming, but I believe we first need some more or less stable api agreed upon. Design doc is very good and very insightful, but at the same time sometimes leaves some (devil in) details unspecified, which might affect result interfaces/implementation quite significantly (like we already discussing about error handling strategies).

include/libcyphal/transport/can/media.hpp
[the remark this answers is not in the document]
serges147: I will add `cetl::optional<MediaError>`.
`do_not_coalesce` will be addressed by issue #348

include/libcyphal/transport/can/transport.hpp
serges147: Please let me fix this in separate dedicated PR - I have already 3 work-in-progress private branches. Changing this right now will make me conflicts to resolve in all these WIP branches. As a sure thing promise here is explicit issue: https://github.com/OpenCyphal-Garage/libcyphal/issues/349

include/libcyphal/transport/can/media.hpp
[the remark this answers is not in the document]
pavel-kirienko: What Scott called `MediaError` is currently named `PlatformError` in the design doc and the codebase. We need to either use `PlatformError` or rename it into `MediaError`.

pavel-kirienko (approved):

include/libcyphal/transport/msg_sessions.hpp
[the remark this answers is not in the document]
thirtytwobits: It's mostly the class-level documentation that I'm struggling with. I'm left to linking your intentions in these PRs with a single word – the class name – against the design document. If you have to c&p from that document it's fine just help me out with some context when you first define a class, interface, or data structure.

include/libcyphal/transport/can/media.hpp
thirtytwobits: Does it have to affect the lizards immediately? My thought is, as long as we establish that the media layer supplies memory which the higher layers are simply borrowing we are free to implement this against the lizards in various ways (the simplest being the current copy into their memory space we are setting up in our media layer apis just behind the media abstraction). This defers such optimizations to the underlying implementation of the media layer and supports various schemes. Am I being naive here?

include/libcyphal/transport/can/transport.hpp
thirtytwobits: ok

include/libcyphal/transport/defines.hpp
thirtytwobits: ok

include/libcyphal/runnable.hpp
```diff
@@ -0,0 +1,31 @@
...
+    IRunnable(IRunnable&&)                 = delete;
+    IRunnable(const IRunnable&)            = delete;
+    IRunnable& operator=(IRunnable&&)      = delete;
+    IRunnable& operator=(const IRunnable&) = delete;
+
+    virtual void run(const TimePoint now) = 0;
+
+protected:
+    IRunnable()          = default;
+    virtual ~IRunnable() = default;
```
thirtytwobits, line 26: It's sorta a nitpick but per [Core Cpp Guidelines C35](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c35-a-base-class-destructor-should-be-either-public-and-virtual-or-protected-and-non-virtual) the `~IRunnable()` shouldn't be virtual. Can we establish this as a standard in this codebase too?

include/libcyphal/transport/can/media.hpp
```diff
@@ -0,0 +1,88 @@
...
+};
+
+class IMedia
+{
+public:
+    IMedia(IMedia&&)                 = delete;
+    IMedia(const IMedia&)            = delete;
+    IMedia& operator=(IMedia&&)      = delete;
+    IMedia& operator=(const IMedia&) = delete;
+    virtual ~IMedia()                = default;
```
thirtytwobits, line 43: protected/non-virtual

include/libcyphal/transport/defines.hpp
```diff
@@ -0,0 +1,90 @@
+/// @copyright
+/// Copyright (C) OpenCyphal Development Team  <opencyphal.org>
+/// Copyright Amazon.com Inc. or its affiliates.
+/// SPDX-License-Identifier: MIT
+
+#ifndef LIBCYPHAL_TRANSPORT_DEFINES_HPP_INCLUDED
```
thirtytwobits, line 6: can we call this `types.hpp` instead? We specifically do not want `#define`s

include/libcyphal/transport/multiplexer.hpp
```diff
@@ -0,0 +1,33 @@
...
+{
+
+class IMultiplexer
+{
+public:
+    IMultiplexer(IMultiplexer&&)                 = delete;
+    IMultiplexer(const IMultiplexer&)            = delete;
+    IMultiplexer& operator=(IMultiplexer&&)      = delete;
+    IMultiplexer& operator=(const IMultiplexer&) = delete;
+    virtual ~IMultiplexer()                      = default;
```
thirtytwobits, line 21: Same here. Interface destructors should always be protected/non-virtual

include/libcyphal/transport/udp/transport.hpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+    CETL_NODISCARD Expected<UniquePtr<IResponseTxSession>, AnyError> makeResponseTxSession(
+        const ResponseTxParams&) override
+    {
+        return NotImplementedError{};
+    }
+
+    // MARK: IRunnable
+
+    void run(const TimePoint) override {}
```
thirtytwobits, line 75: prefer final to override. If actually using override document how subclasses must handle the base-class implementation (e.g. they shall-or-shall-not invoke it. They shall-invoke it first/last. etc).

thirtytwobits (commented):
Generally all good here. Please tighten up the interface destructors as this can be hard to fix as the design evolved.

The biggest issue is the open issue on media-layer memory ownership. If we need a interim design call I can do that.

include/libcyphal/runnable.hpp
serges147: It's a bit tricky (for me at least) to decide when we want destructor accessible and when we don't... Even for interfaces we currently do move them into unique_ptr<Concrete> → unique_ptr<Interface> which will require them public.

include/libcyphal/transport/multiplexer.hpp
serges147: I get it for external entity interfaces (like media and multiplexer), but for our own entities we implement (and return them via `makeXxx` as unique ptrs) it still blurry which pattern to use.

include/libcyphal/transport/udp/transport.hpp
serges147: At class level we have `final` - afaik it means that all the virtual methods are final as well. do you want make it even more explicit?

serges147: About "shall-or-shall-not invoke" I can't comment yet (due to lack of fine understanding of requirements), but I agree we should have (eventually) such documentation.

include/libcyphal/transport/dynamic_buffer.hpp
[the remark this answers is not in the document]
pavel-kirienko: How about `FragmentedBuffer` or `ScatteredBuffer`?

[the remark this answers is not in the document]
pavel-kirienko: How about `FragmentedBuffer` or `ScatteredBuffer`?

include/libcyphal/runnable.hpp
thirtytwobits: We shouldn't manage the lifecycle of an object through it's pointer. This can lead to dangerous assumptions and amputations (https://godbolt.org/z/1Tj4nzehd). This is what the ~`voldemort_ptr`~, ~`dark-ptr`~, `interface_ptr` type is meant to solve.

include/libcyphal/transport/dynamic_buffer.hpp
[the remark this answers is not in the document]
thirtytwobits: Fragmented is a dirty word so let's go with ScatteredBrain, er, ScatteredBuffer

include/libcyphal/transport/udp/transport.hpp
thirtytwobits: Eh. I guess if the class is final the sure. It's fine.

include/libcyphal/transport/dynamic_buffer.hpp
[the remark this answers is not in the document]
serges147: I'll change it to →  `ScatteredBuffer`

include/libcyphal/transport/defines.hpp
serges147: I will rename it in the next (347) pr.

include/libcyphal/transport/dynamic_buffer.hpp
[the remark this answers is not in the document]
serges147: done

include/libcyphal/runnable.hpp
serges147: Thank You Scott for very representable example (with legless dog). Please see latest commit where I moved destructors for "external" entities (`IMedia`-s and `IMultiplexer`) to protected area. BUT, I can't currently do the same for `IRunnable` up to various `IXxxSession`-s we return to user as unique ptrs. This is b/c of our current pmr implementation of UniquePtr - it REQUIRES currently that ~Interface will be available for deleter. The only way I see to remove such requirement is to have type-erased deleter which will capture in a closure original concrete type allocator. It will require `cetl::function` (not yet available). IMHO it has to be type-erased - otherwise we can't have proper `UniquePtr<IX>` → `UniquePtr<IY>` conversions (assuming of course that `X` and `Y` convertible). I discussed this with Pavel - it seems that he agrees with me.

BTW, such type-erased deleter should also solve current standing issue https://github.com/OpenCyphal/CETL/issues/118 (see 2nd bullet)

include/libcyphal/transport/can/transport.hpp
serges147: It looks like this in CLion:
Screenshot 2024-04-23 at 12 46 01

include/libcyphal/runnable.hpp
thirtytwobits: Okay. Let's track this with TODOs in the code. Perhaps I'll take on this work in CETL. It'll give me a reason to move my godbolt example into github.

include/libcyphal/transport/multiplexer.hpp
emrainey: Internal implementations would not be destructed via the interface though? They would use their complete internal types.

thirtytwobits (approved):
As discussed on call: we have the right issues captured now to ensure my overarching concerns will be addressed moving forward

include/libcyphal/transport/can/media.hpp
pavel-kirienko: Moved into https://github.com/OpenCyphal-Garage/libcyphal/issues/352

include/libcyphal/transport/scattered_buffer.hpp
```diff
@@ -0,0 +1,120 @@
...
+        CETL_NODISCARD virtual std::size_t size() const noexcept                      = 0;
+        CETL_NODISCARD virtual std::size_t copy(const std::size_t offset_bytes,
+                                                void* const       destination,
+                                                const std::size_t length_bytes) const = 0;
+
+    protected:
+        Interface()                                = default;
+        Interface(Interface&&) noexcept            = default;
+        Interface& operator=(Interface&&) noexcept = default;
+
```
pavel-kirienko, line 45: the rule of 5 requires the dtor to be explicitly defaulted

pavel-kirienko (approved):

include/libcyphal/transport/scattered_buffer.hpp
serges147: will do it at after-next pr

Issue #226, opened 2019-05-26
integrate clang-tidy with v1 build


thirtytwobits:
run clang-tidy at build-time.

pavel-kirienko:
Clang-Tidy should be integrated via `CXX_CLANG_TIDY` and Sonar needs to be invoked on pull request builds or when the commit is tagged with `#sonar`.

@serges147 I would like to prioritize this because the way the codebase is now we're already going to have to go back and modify it to make Clang-Tidy and Sonar accept it.

serges147:
@pavel-kirienko Pavel, I'm right that I can use exactly the same approach as in CETL in terms of Sonar? Also where to get `clang-tidy` config file?

pavel-kirienko:
1. Yes; alternatively, you could use Dyshlo as a reference.
2. Let's start with this one (it may require some tweaking but we'll get to that later):

```yaml
Checks: >-
  boost-*,
  bugprone-*,
  cert-*,
  clang-analyzer-*,
  cppcoreguidelines-*,
  google-*,
  hicpp-*,
  llvm-*,
  misc-*,
  modernize-*,
  performance-*,
  portability-*,
  readability-*,
  -cppcoreguidelines-avoid-const-or-ref-data-members,
  -google-readability-todo,
  -readability-avoid-const-params-in-decls,
  -readability-identifier-length,
  -llvm-header-guard,
  -llvm-include-order,
  -*-use-trailing-return-type,
  -*-named-parameter,
CheckOptions:
  - key:   readability-function-cognitive-complexity.Threshold
    value: '90'
  - key:   readability-magic-numbers.IgnoredIntegerValues
    value: '1;2;3;4;5;8;10;16;20;32;60;64;100;128;256;500;512;1000'
WarningsAsErrors: '*'
HeaderFilterRegex: '.*\.hpp'
AnalyzeTemporaryDtors: false
FormatStyle: file
```

PR #356, opened 2024-05-02
Service RX/TX Request/Response sessions

serges147:
- Added Svc RX & TX sessions (both for Request & Response).
- Now that we have more sessions, `ContiguousPayload` building moved to transport impl:
  - less code duplication
  - less to verify by unit tests
  - sessions now more like facade for RX/TX - they do necessary params validation/conversion, but ultimately delegate to the transport
- switch to `cetl::unbounded_variant` (instead of former `cetl::any`)

Also:
- minor changes for previous PR requests
- more docs for stable api; excluding `*::detail::*` from generated docs

Codo coverage:
Screenshot 2024-05-02 at 09 53 11


.clang-tidy
```diff
@@ -0,0 +1,31 @@
...
+  -*-named-parameter,
+CheckOptions:
+  - key:   readability-function-cognitive-complexity.Threshold
+    value: '90'
+  - key:   readability-magic-numbers.IgnoredIntegerValues
+    value: '1;2;3;4;5;8;10;16;20;32;60;64;100;128;256;500;512;1000'
+WarningsAsErrors: '*'
+HeaderFilterRegex: '.*\.hpp'
+AnalyzeTemporaryDtors: false
+FormatStyle: file
```
serges147, line 31: Not yet integrated, but just initial version of the config according issue #226

cmake/modules/Findcetl.cmake
```diff
@@ -6,7 +6,7 @@
 
 include(FetchContent)
 set(cetl_GIT_REPOSITORY "https://github.com/OpenCyphal/cetl.git")
-set(cetl_GIT_TAG "886a0d227a043511eed6b252ea0f788590c50e75")
+set(cetl_GIT_TAG "10fbb2b7b89473d68e73db7235848b0692169e5a")
```
serges147, line 9: CETL's `main` (with recently merged `cetl::unbounded_variant`.

docs/doxygen.ini
```diff
@@ -1071,7 +1071,7 @@ EXCLUDE_PATTERNS       =
 # wildcard * is used, a substring. Examples: ANamespace, AClass,
 # ANamespace::AClass, ANamespace::*Test
 
-EXCLUDE_SYMBOLS        =
+EXCLUDE_SYMBOLS        = *::detail::*
```
serges147, line 1074: Internal implementation detail won't go to generated docs.

include/libcyphal/transport/can/delegate.hpp
```diff
@@ -230,26 +230,28 @@ class TransportDelegate
...
 
-/// This internal session delegate class serves the following purpose: it provides interface
-/// to access a session from transport (by casting canard's `user_reference` member to this class).
+// MARK: -
+
+/// This internal session delegate class serves the following purpose: it provides an interface (aka gateway)
+/// to access RX session from transport (by casting canard's `user_reference` member to this class).
 ///
-class SessionDelegate
+class IRxSessionDelegate
```
serges147, line 238: Adding `I` prefix for anything like an interface (regardless whether it's internal or public stuff) - as it was requested in previous PR.

include/libcyphal/transport/can/msg_rx_session.hpp
```diff
@@ -25,22 +25,33 @@ namespace can
...
 
 public:
     CETL_NODISCARD static Expected<UniquePtr<IMessageRxSession>, AnyError> make(TransportDelegate&     delegate,
                                                                                 const MessageRxParams& params)
     {
-        auto session = libcyphal::detail::makeUniquePtr<Tag>(delegate.memory(), Tag{}, delegate, params);
+        if (params.subject_id > CANARD_SUBJECT_ID_MAX)
+        {
+            return ArgumentError{};
+        }
```
serges147, line 52: Parameters are const (see `params_` below), and their `subject_id` won't change. So, it makes sense fail early (at `make` phase) rather than later during transmissions. The same story for other session factories.

include/libcyphal/transport/can/svc_rx_sessions.hpp
```diff
@@ -0,0 +1,169 @@
...
+///
+/// @tparam Interface_ Type of the session interface.
+///                    Could be either `IRequestRxSession` or `IResponseRxSession`.
+/// @tparam Params Type of the session parameters.
+///                Could be either `RequestRxParams` or `ResponseRxParams`.
+/// @tparam TransferKind Kind of the service transfer.
+///                      Could be either `CanardTransferKindRequest` or `CanardTransferKindResponse`.
+///
+template <typename Interface_, typename Params, CanardTransferKind TransferKind>
+class SvcRxSession final : public Interface_, private IRxSessionDelegate
```
serges147, line 39: Both service RX sessions (rx request and rx response) are almost the same in terms of structure and functionality. The only slight difference is their interface name, params type and "kind" of canard transfer. So, IMO making template here is fine and justifiable b/c ...:
1. it's `detail`-s, not exposed to a user
2. no code duplication
3. both instantiations (see below just after the template) are 100% covered by unit tests (at line, *function* and *branches* level).

include/libcyphal/transport/can/svc_rx_sessions.hpp
```diff
@@ -0,0 +1,169 @@
...
+    {
+        const int8_t result = ::canardRxSubscribe(&delegate.canard_instance(),
+                                                  TransferKind,
+                                                  static_cast<CanardPortID>(params_.service_id),
+                                                  static_cast<std::size_t>(params_.extent_bytes),
+                                                  CANARD_DEFAULT_TRANSFER_ID_TIMEOUT_USEC,
+                                                  &subscription_);
+        (void) result;
+        CETL_DEBUG_ASSERT(result >= 0, "There is no way currently to get an error here.");
+        CETL_DEBUG_ASSERT(result > 0, "New subscription supposed to be made.");
```
serges147, line 85: There is no currently path to fail - pre-validation done at factory.

include/libcyphal/transport/can/svc_rx_sessions.hpp
serges147: BTW, `MessageRxSession` is also very similar, BUT still has its differences (in terms of params validation and anonymous case handling), so it does not fit into such RX template IMO (hence there is minor code duplication - all covered of course by tests).

include/libcyphal/transport/can/svc_rx_sessions.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+        subscription_.user_reference = static_cast<IRxSessionDelegate*>(this);
+    }
+
+    ~SvcRxSession() final
+    {
+        ::canardRxUnsubscribe(&delegate_.canard_instance(),
+                              TransferKind,
+                              static_cast<CanardPortID>(params_.service_id));
```
serges147, line 94: No path fail as well, but if you think its worth to assert I can do it (here and for msg rx session).

include/libcyphal/transport/scattered_buffer.hpp
```diff
@@ -15,53 +15,99 @@ namespace libcyphal
...
+    /// @brief Defines maximum size (aka footprint) of the storage variant.
+    ///
+    static constexpr std::size_t StorageVariantFootprint = sizeof(void*) * 8;
 
-    class Interface : public cetl::rtti_helper<InterfaceTypeIdType>
+    /// @brief Defines storage interface for the scattered buffer.
+    ///
+    /// @see ScatteredBuffer::ScatteredBuffer(AnyStorage&& any_storage)
+    ///
+    class IStorage : public cetl::rtti_helper<IStorageTypeIdType>
```
serges147, line 38: Here and there (including in the design doc) we use word "storage" when we talk about Lizard-specific implementation of actual (potentially) scattered storage this "facade" `ScatteredBuffer` is using. So, I decided to rename nested `Interface` to IMO more meaningful `IStorage`.

test/unittest/gtest_helpers.hpp
```diff
@@ -67,8 +67,8 @@ namespace libcyphal
 
 inline void PrintTo(const Duration duration, std::ostream* os)
 {
-    auto locale = os->imbue(std::locale("en_US"));
-    *os << std::chrono::duration_cast<std::chrono::microseconds>(duration).count() << "_us";
+    auto locale = os->imbue(std::locale(""));
+    *os << std::chrono::duration_cast<std::chrono::microseconds>(duration).count() << " us";
```
serges147, line 71: Previous version failed on my "Estonian only" Ubuntu (b/c there is no "en_US" locale there) ;)
BTW, locale stuff here is just to have more human friendly gtest failure output (like `1'000'000 us` vs `1000000 us` without it).

test/unittest/gtest_helpers.hpp
```diff
@@ -252,6 +288,42 @@ inline testing::Matcher<const CanId&> SourceNodeOfCanIdEq(NodeId node_id)
...
+    }
+
+private:
+    const NodeId node_id_;
+};
+inline testing::Matcher<const CanId&> DestinationNodeOfCanIdEq(NodeId node_id)
+{
+    return DestinationNodeCanIdMatcher(node_id);
+}
+
```
serges147, line 326: All these gmock matchers (and `PrintTo` helpers) have no use when all the tests are green (b/c there is no output), but it makes big difference when something goes wrong. Using them (in conjunction with `EXPECT_THAT` macro) gives you nice print of what was expected and what actually is. Stay tuned, even more will be added...

include/libcyphal/transport/can/svc_rx_sessions.hpp
thirtytwobits: Sorry, why can't this fail?

include/libcyphal/transport/can/svc_rx_sessions.hpp
thirtytwobits: I like the debug asserts as long as we are able to verify them.

include/libcyphal/transport/can/svc_rx_sessions.hpp
```diff
@@ -0,0 +1,169 @@
...
+        {
+            subscription_.transfer_id_timeout_usec = static_cast<CanardMicrosecond>(timeout_us.count());
+        }
+    }
+
+    // MARK: IRunnable
+
+    void run(const TimePoint) final
+    {
+        // Nothing to do here currently.
```
thirtytwobits, line 127: image

include/libcyphal/transport/udp/media.hpp
```diff
@@ -25,7 +27,12 @@ class IMedia
     IMedia& operator=(const IMedia&)     = delete;
     IMedia& operator=(IMedia&&) noexcept = delete;
 
-    // TODO: Add methods here
+    /// @brief Get the maximum transmission unit (MTU) of the UDP media.
+    ///
+    /// This value may change arbitrarily at runtime. The transport implementation will query it before every
+    /// transmission on the port. This value has no effect on the reception pipeline as it can accept arbitrary MTU.
+    ///
+    CETL_NODISCARD virtual std::size_t getMtu() const noexcept = 0;
```
thirtytwobits, line 35: This is a public API and nothing catastrophic happens if the value isn't used. Please omit `CETL_NODISCARD` value

thirtytwobits (changes requested):

include/libcyphal/transport/can/svc_rx_sessions.hpp
pavel-kirienko: @thirtytwobits the only possible error condition occurs if we feed invalid arguments, which we pre-validate and know are correct; hence, it can't fail.

include/libcyphal/transport/udp/media.hpp
serges147: ok, I will do it. But there is separate issue about revising everything in terms of "to discard or not to discard" question.

thirtytwobits (approved):

PR #357, opened 2024-05-06
CAN media filters reconfiguration

serges147:
- Implemented CAN media filters reconfiguration on subscribe/unsubscribe
- Reference official `libcanard` v3.2.0 #verification


include/libcyphal/transport/can/delegate.hpp
```diff
@@ -166,6 +206,15 @@ class TransportDelegate
...
                                                                  const PayloadFragments        payload_fragments) = 0;
 
+    /// @brief Triggers update of media filters due to a change of an RX subscription.
+    ///
+    /// Actual update will be done on next `run` of transport.
+    ///
+    /// @param is_service `true` if called from a service subscription; `false` for a message one.
+    /// @param is_subscription_added Indicates whether a new subscription was added; otherwise, it was removed.
+    ///
+    virtual void triggerUpdateOfFilters(const bool is_service, const bool is_subscription_added) noexcept = 0;
```
pavel-kirienko, line 216: Can we please use strong types here instead of the two booleans? At the point of invocation, the two booleans back to back look confusing.

Also, I think I might have mentioned this earlier: `subscription` is not a very good choice of word here. I know that it's coming from libcanard but that's a separate issue. A more appropriate term is "port", can we use that instead?

include/libcyphal/transport/can/transport.hpp
```diff
@@ -442,9 +480,110 @@ class TransportImpl final : public ICanTransport, private TransportDelegate
...
+
+            filters_span = {filters.data(), filters.size()};
+        }
+
+        // Let know each media interface about the new filters (tracking the fact of possible media error).
+        //
+        bool was_error = false;
+        for (const Media& media : media_array_)
+        {
+            const cetl::optional<MediaError> maybe_error = media.interface.setFilters(filters_span);
```
pavel-kirienko, line 567: I don't understand what do we need the `filters_span` for? Can't we just construct it as a temporary ad-hoc instead of introducing a new named entity? It should be implicitly constructible from the array.

```suggestion
            const cetl::optional<MediaError> maybe_error = media.interface.setFilters(filters);
```

include/libcyphal/transport/can/delegate.hpp
```diff
@@ -104,6 +105,45 @@ class TransportDelegate
...
+    /// @tparam Node Type of a concrete AVL tree node.
+    ///
+    template <typename Node>
+    struct CanardConcreteTree
+    {
+        /// @brief Recursively visits each node of the AVL tree, counting total number of nodes visited.
+        ///
+        /// Recursion goes first to the left child, then to the current node, and finally to the right child.
+        /// B/c AVL tree is balanced, the total complexity is `O(n)` and call stack depth
+        /// should not be deeper than `O(log(N))` (max 13 in case of 8192 active message subscriptions),
```
pavel-kirienko, line 119: Nit but the maximum depth of an AVL tree is $\lt \lceil{} 1.44 \log_2(N+2) - 0.328\rceil{}$, which yields <19 call stack levels

include/libcyphal/transport/can/delegate.hpp
```diff
@@ -104,6 +105,45 @@ class TransportDelegate
...
+            if (node == nullptr)
+            {
+                return 0;
+            }
+
+            // Initial `1` is for the current node.
+            std::size_t count = 1;
+
+            count += visitCounting(node->lr[0], std::forward<Visitor>(visitor));
+            visitor(*reinterpret_cast<Node*>(node));
```
pavel-kirienko, line 139: It's worth adding that the reinterpret cast is unavoidable because we have to deal with a low-level and unsafe C API.

pavel-kirienko (approved):
I left a few low-severity comments

include/libcyphal/transport/can/transport.hpp
serges147: No we can't:
```
/Users/sergei/Develop/git/OpenCyphal-Garage/libcyphal/include/libcyphal/transport/can/transport.hpp:567:87: error: no viable conversion from 'libcyphal::detail::VarArray<Filter>' (aka 'VariableLengthArray<libcyphal::transport::can::Filter, polymorphic_allocator<libcyphal::transport::can::Filter>>') to 'Filters' (aka 'span<const libcyphal::transport::can::Filter, 18446744073709551615UL>')
            const cetl::optional<MediaError> maybe_error = media.interface.setFilters(filters);
                                                                                      ^~~~~~~
```

, but we you right about not needed `filters_span`: we can do like this:
```
const cetl::optional<MediaError> maybe_error = media.interface.setFilters({filters.data(), filters.size()});
```

include/libcyphal/transport/can/delegate.hpp
pavel-kirienko: Yeah that's one way to do it. In C++20 we would just introduce a struct but there's no way to sensibly initialize it in the earlier versions of the standard so let's go with an enum I guess.

include/libcyphal/transport/can/delegate.hpp
serges147: thanks, good to know - I will correct it.

include/libcyphal/transport/can/delegate.hpp
serges147: It's actually in my next PR - where I added clang_tidy.

pavel-kirienko (approved):
We'll need to have a serious talk about Sonar

include/libcyphal/transport/can/transport.hpp
```diff
@@ -442,9 +485,107 @@ class TransportImpl final : public ICanTransport, private TransportDelegate
         }  // for each media
     }
 
+    /// \brief Runs (if needed) reconfiguration of media filters based on the currently active subscriptions.
+    ///
+    /// Temporary allocates memory buffers for all filters, one per each active subscription (message or service).
+    /// In case of redundant media, each media interface will be called with the same span of filters.
+    /// In case of zero subscriptions, we still need to call media interfaces to clear their filters,
+    /// through there will be no memory allocation for the empty buffer.
```
lydia-at-amazon, line 493: nit: `through` --> `though`?

include/libcyphal/transport/can/transport.hpp
```diff
@@ -442,9 +485,107 @@ class TransportImpl final : public ICanTransport, private TransportDelegate
...
+    ///
+    /// @note Service RX subscriptions are not considered as active ones for \b anonymous nodes.
+    ///
+    /// @note If \b whole reconfiguration process was successful,
+    /// `should_reconfigure_filters_` will be reset to `false`, so that next time the run won't do any job.
```
lydia-at-amazon, line 498: nit: `won't do any job` --> `won't do any work`?

include/libcyphal/transport/can/transport.hpp
```diff
@@ -442,9 +485,107 @@ class TransportImpl final : public ICanTransport, private TransportDelegate
...
+                };
+                ports_count += RxSubscriptionTree::visitCounting(subs_trees[CanardTransferKindRequest], svc_visitor);
+                ports_count += RxSubscriptionTree::visitCounting(subs_trees[CanardTransferKindResponse], svc_visitor);
+            }
+
+            (void) ports_count;
+            CETL_DEBUG_ASSERT(ports_count == total_active_ports, "");
+        }
+
+        // Let know each media interface about the new filters (tracking the fact of possible media error).
```
lydia-at-amazon, line 564: nit: `Let know each media interface` --> `Let each media interface know`

test/unittest/transport/can/test_can_delegate.cpp
```diff
@@ -176,4 +177,65 @@ TEST_F(TestCanDelegate, canardMemoryAllocate_no_memory)
...
+        {
+        }
+        std::string name;
+    };
+    //        Root
+    //      ↙     ↘
+    //  Left       Right
+    //      ↘      ↙   ↘
+    //       LR  RL     RR
+    //
```
lydia-at-amazon, line 196: Thank you for the diagram :)

lydia-at-amazon (approved):

lydia-at-amazon:
What does #verification mean?

pavel-kirienko:
@lydia-at-amazon see https://github.com/OpenCyphal-Garage/libcyphal/blob/main/CONTRIBUTING.md#issue-and-hashtag-based-ci-triggering

PR #358, opened 2024-05-09
Clang tidy & sonar integration and corresponding fixes.

serges147:
GitHub's "Hide whitespace" is recommended during review.

image

image


include/libcyphal/transport/can/delegate.hpp
```diff
@@ -61,7 +72,7 @@ class TransportDelegate
         }
         CanardMemory(const CanardMemory&) = delete;
 
-        ~CanardMemory() final
+        ~CanardMemory() override
```
serges147, line 75: clang-tidy wants like this, so I reverted back to `override` everywhere

include/libcyphal/transport/udp/transport.hpp
```diff
@@ -46,7 +55,7 @@ class TransportImpl final : public IUdpTransport
     TransportImpl(Spec,
                   cetl::pmr::memory_resource&            memory,
                   IMultiplexer&                          multiplexer,
-                  libcyphal::detail::VarArray<IMedia*>&& media_array,
+                  libcyphal::detail::VarArray<IMedia*>&& media_array,  // NOLINT
```
serges147, line 58: this is temporary - there is no implementation (yet), so no `std::move` of the `media_array`).

include/libcyphal/transport/can/delegate.hpp
```diff
@@ -6,14 +6,21 @@
...
+#include "libcyphal/transport/errors.hpp"
+#include "libcyphal/transport/scattered_buffer.hpp"
+#include "libcyphal/transport/types.hpp"
+#include "libcyphal/types.hpp"
 
 #include <canard.h>
+#include <cetl/cetl.hpp>
+#include <cetl/pf17/attribute.hpp>
+#include <cetl/pf17/cetlpf.hpp>
 #include <cetl/rtti.hpp>
```
lydia-at-amazon, line 18: nit: These are not technically system headers, so should we be using `""` instead of `<>`? We have been discussing this internally as well. And are leaning towards always using `""`. I know it's common to use `<>` for any headers that aren't part of the local repo, but it tells the compiler to first search the system headers and it won't find these headers there.

serges147: These are all done like this intentionally and I believe quite consistently - everything which is not directly part of libcyphal is treated as "system" (or external). In unit tests I'm going even further - libcyphal itself is treated there like "external" stuff (kinda emulating how clients of our library supposed to use it). Also, AFAIK, `""` have "relative path notation" which is not applicable for any external stuff (whether it's a real system header or just dependency to a submodule like CETL).

.clang-tidy
```diff
@@ -12,12 +12,17 @@ Checks: >-
...
   readability-*,
+  -clang-analyzer-core.uninitialized.Assign,
   -cppcoreguidelines-avoid-const-or-ref-data-members,
+  -google-readability-avoid-underscore-in-googletest-name,
   -google-readability-todo,
+  -llvm-header-guard,
+  -modernize-concat-nested-namespaces,
+  -modernize-type-traits,
+  -modernize-use-constraints,
+  -modernize-use-nodiscard,
```
pavel-kirienko, line 23: Shouldn't we be using our `CETL_NODISCARD` where it says nodiscard is needed?

include/libcyphal/transport/can/delegate.hpp
```diff
@@ -152,13 +170,20 @@ class TransportDelegate
         ServicePortRemoved
     };
 
+    // NOSONAR below b/c Sonar's cpp:S134 conflicts with AUTOSAR A12-1-2
```
pavel-kirienko, line 173: Conflicting rules should be turned off in Sonar

include/libcyphal/transport/can/msg_rx_session.hpp
```diff
@@ -63,13 +76,12 @@ class MessageRxSession final : public IMessageRxSession, private IRxSessionDeleg
     MessageRxSession(Spec, TransportDelegate& delegate, const MessageRxParams& params)
         : delegate_{delegate}
         , params_{params}
-        , subscription_{}
-        , last_rx_transfer_{}
+        , subscription_{}  // NOSONAR : Sonar's cpp:S134 conflicts with AUTOSAR A12-1-2
```
pavel-kirienko, line 79: Same here; I will omit further comments on the same deficiency.

include/libcyphal/transport/udp/transport.hpp
pavel-kirienko: It is best to comment on that directly in code. For example, a good practice is to always provide a very short explanation right after `NOSONAR`. We even have a script to enforce that: https://github.com/Zubax/dyshlo/blob/main/tools/check_nolint.sh

test/unittest/CMakeLists.txt
```diff
@@ -60,25 +65,37 @@ foreach(NATIVE_TEST ${NATIVE_TESTS})
...
 )
+
+if (CMAKE_BUILD_TYPE STREQUAL "Coverage")
+
+    enable_coverage_report(COVERAGE_REPORT_FORMATS html sonarqube
+        ROOT_DIRECTORY ${LIBCYPHAL_ROOT}
+        OUT_REPORT_INDICIES LOCAL_COVERAGE_REPORT_INDICIES
+    )
+
+    # we need to build at least one source file to get sonarqube to look at the cetl
```
pavel-kirienko, line 91: why cetl

pavel-kirienko: You may want to know that there's another way to use Sonar with a header-only library: just analyze your test build, and configure exclusion rules to exclude all .cpp files, as done here: https://github.com/Zubax/dyshlo/blob/27e0f89970e476073eef9c9a556d74a19c632da2/tools/run_sonar.sh#L59-L74

test/unittest/transport/can/test_can_delegate.cpp
```diff
@@ -3,18 +3,32 @@
...
+#include <utility>
+#include <vector>
 
 namespace
 {
-using namespace libcyphal::transport;
-using namespace libcyphal::transport::can;
+
+using libcyphal::TimePoint;
+using namespace libcyphal::transport;  // NOLINT This our main concern here in the unit tests.
```
pavel-kirienko, line 31: It is best to provide NOLINT comments with a specific rule to be suppressed. CLion can inject these automatically.

pavel-kirienko (commented):
I wonder if we should disable the no-magic-numbers rules? Magic numbers may be a problem in numerical code but this is not the case here. I don't like the suppression comments near the type_id_type definitions.

.clang-tidy
serges147: `CETL_NODISCARD` is not working for c++14 - this rule expects specifically `[[nodiscard]]`

include/libcyphal/transport/can/delegate.hpp
serges147: They are not always conflicting.

serges147:
> I wonder if we should disable the no-magic-numbers rules? Magic numbers may be a problem in numerical code but this is not the case here. I don't like the suppression comments near the type_id_type definitions.

I disabled them at unit tests. Production code still under this "no-magic-numbers" rule with following exceptions in the `.clang-tidy` file:
```
  - key:   readability-magic-numbers.IgnoredIntegerValues
    value: '1;2;3;4;5;8;10;16;20;32;60;64;100;128;256;500;512;1000'
```

include/libcyphal/transport/can/delegate.hpp
serges147: OK, I agree with "erode" theme - I'll try disable this tidy rule.

include/libcyphal/transport/can/delegate.hpp
pavel-kirienko: I am with Sergei here because per the standard, the behavior of `""` is a superset of `<>` -- if the former fails, it falls back to `<>`; if we know that it is going to fail there is no utility in using it.

image

pavel-kirienko (approved):

include/libcyphal/transport/can/delegate.hpp
lydia-at-amazon: Thank you! I am aligned with continuing to use `<>` for "external" stuff.

@thirtytwobits FYI since we were just discussing this the other day.

lydia-at-amazon (approved):

Issue #350, opened 2024-04-24
Adjust the use of CETL_NODISCARD


pavel-kirienko:
- All non-void internal functions can (should) continue using CETL_NODISCARD.

- User-facing (API) functions can only use CETL_NODISCARD where unused return value may lead to a resource leak or some other disaster.

serges147:
Closing as it was addressed by [PR 359](https://github.com/OpenCyphal-Garage/libcyphal/pull/359)

PR #359, opened 2024-05-13
No CETL_NODISCARD for most of the public api

serges147:
No `CETL_NODISCARD` for the most of public api (issue #350), but private (aka `detail`) stuff stays under strict `[[nodiscard]]` policy for all non-`void` returning functions.

Also:
- Less `NOSONAR`-s by switching some internal `void*`-s → `cetl::byte*`-s.
- Eliminated `IMultiplexer` at CAN - no use of it; but will be at UDP.
- More docs for public api.
- Fixed Build Status and Sonar Cloud badges at `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 ![Cyphal](docs/images/html/opencyphal_logo.svg) Cyphal stack in C++
 ===================
 
-[![Build Status](https://badge.buildkite.com/af844974c06af6406e3b2192d98298b02b30f6ebebb5f8b16c.svg)](https://buildkite.com/uavcan/libcyphal-v1)
+[![Build Status](https://github.com/OpenCyphal-Garage/libcyphal/actions/workflows/tests.yml/badge.svg)](https://github.com/OpenCyphal-Garage/libcyphal)
 [![Forum](https://img.shields.io/discourse/https/forum.opencyphal.org/users.svg)](https://forum.opencyphal.org)
-[![Sonarqube Badge](https://sonarcloud.io/api/project_badges/measure?project=UAVCAN_libuavcan&metric=alert_status)](https://sonarcloud.io/dashboard?id=UAVCAN_libuavcan)
-[![Sonarqube Coverage](https://sonarcloud.io/api/project_badges/measure?project=UAVCAN_libuavcan&metric=coverage)](https://sonarcloud.io/dashboard?id=UAVCAN_libuavcan)
+[![Sonarqube Badge](https://sonarcloud.io/api/project_badges/measure?project=OpenCyphal-Garage_libcyphal&metric=alert_status)](https://sonarcloud.io/project/overview?id=OpenCyphal-Garage_libcyphal)
+[![Sonarqube Coverage](https://sonarcloud.io/api/project_badges/measure?project=OpenCyphal-Garage_libcyphal&metric=coverage)](https://sonarcloud.io/project/overview?id=OpenCyphal-Garage_libcyphal)
 [![Documentation](https://img.shields.io/badge/docs-passing-green.svg)](https://opencyphal.org/libcyphal/)
 
 > **WARNING** libcyphal v1 is not yet complete. This is a work-in-progress.
@@ -12,4 +12,3 @@
 Portable reference implementation of the [Cyphal protocol stack](https://opencyphal.org) in C++ for embedded systems, Linux, and POSIX-compliant RTOSs.
 
 Cyphal is a lightweight protocol designed for reliable communication in aerospace and robotic applications over robust vehicular networks.
-
--- a/include/libcyphal/transport/can/media.hpp
+++ b/include/libcyphal/transport/can/media.hpp
@@ -10,7 +10,6 @@
 #include "libcyphal/types.hpp"
 
 #include <cetl/cetl.hpp>
-#include <cetl/pf17/attribute.hpp>
 #include <cetl/pf17/cetlpf.hpp>
 #include <cetl/pf20/cetlpf.hpp>
 
--- a/include/libcyphal/transport/can/media.hpp
+++ b/include/libcyphal/transport/can/media.hpp
@@ -68,20 +68,22 @@ public:
     ///
     /// @return Returns `nullopt` on success; otherwise some `MediaError` in case of a low-level error.
     ///
-    CETL_NODISCARD virtual cetl::optional<MediaError> setFilters(const Filters filters) noexcept = 0;
+    virtual cetl::optional<MediaError> setFilters(const Filters filters) noexcept = 0;
 
     /// @brief Schedules the frame for transmission asynchronously and return immediately.
     ///
     /// @return Returns `true` if accepted or already timed out; `false` to try again later.
     ///
-    CETL_NODISCARD virtual Expected<bool, MediaError> push(const TimePoint                    deadline,
-                                                           const CanId                        can_id,
-                                                           const cetl::span<const cetl::byte> payload) noexcept = 0;
+    virtual Expected<bool, MediaError> push(const TimePoint                    deadline,
+                                            const CanId                        can_id,
+                                            const cetl::span<const cetl::byte> payload) noexcept = 0;
 
     /// @brief Takes the next payload fragment (aka CAN frame) from the reception queue unless it's empty.
     ///
     /// @param payload_buffer The payload of the frame will be written into the mutable `payload_buffer` (aka span).
     /// @return Description of a received fragment if available; otherwise an empty optional is returned immediately.
+    ///         `nodiscard` is used to prevent ignoring the return value, which contains not only possible media error,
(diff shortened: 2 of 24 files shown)


README.md
```diff
@@ -1,15 +1,14 @@
...
 ===================
 
-[![Build Status](https://badge.buildkite.com/af844974c06af6406e3b2192d98298b02b30f6ebebb5f8b16c.svg)](https://buildkite.com/uavcan/libcyphal-v1)
+[![Build Status](https://github.com/OpenCyphal-Garage/libcyphal/actions/workflows/tests.yml/badge.svg)](https://github.com/OpenCyphal-Garage/libcyphal)
 [![Forum](https://img.shields.io/discourse/https/forum.opencyphal.org/users.svg)](https://forum.opencyphal.org)
-[![Sonarqube Badge](https://sonarcloud.io/api/project_badges/measure?project=UAVCAN_libuavcan&metric=alert_status)](https://sonarcloud.io/dashboard?id=UAVCAN_libuavcan)
-[![Sonarqube Coverage](https://sonarcloud.io/api/project_badges/measure?project=UAVCAN_libuavcan&metric=coverage)](https://sonarcloud.io/dashboard?id=UAVCAN_libuavcan)
+[![Sonarqube Badge](https://sonarcloud.io/api/project_badges/measure?project=OpenCyphal-Garage_libcyphal&metric=alert_status)](https://sonarcloud.io/project/overview?id=OpenCyphal-Garage_libcyphal)
+[![Sonarqube Coverage](https://sonarcloud.io/api/project_badges/measure?project=OpenCyphal-Garage_libcyphal&metric=coverage)](https://sonarcloud.io/project/overview?id=OpenCyphal-Garage_libcyphal)
```
serges147, line 7: When time comes we of course will quickly search/replace "OpenCyphal-Garage" to → "OpenCyphal" everywhere in this repo.

include/libcyphal/transport/can/media.hpp
```diff
@@ -68,20 +68,22 @@ class IMedia
...
+    virtual Expected<bool, MediaError> push(const TimePoint                    deadline,
+                                            const CanId                        can_id,
+                                            const cetl::span<const cetl::byte> payload) noexcept = 0;
 
     /// @brief Takes the next payload fragment (aka CAN frame) from the reception queue unless it's empty.
     ///
     /// @param payload_buffer The payload of the frame will be written into the mutable `payload_buffer` (aka span).
     /// @return Description of a received fragment if available; otherwise an empty optional is returned immediately.
+    ///         `nodiscard` is used to prevent ignoring the return value, which contains not only possible media error,
+    ///         but also important metadata (like `payload_size` field for further parsing of the result payload).
```
serges147, line 86: IMO this is the only public method which has little sense with discarding its result. All other methods could be used without checking results (f.e. if you don't care about possible errors and etc.) - anyway, no leaks or dangerous behavior will happen.

include/libcyphal/transport/can/transport.hpp
```diff
@@ -654,14 +644,22 @@ class TransportImpl final : public ICanTransport, private TransportDelegate  //
...
-    const cetl::span<IMedia*>    media,
-    const std::size_t            tx_capacity,
-    const cetl::optional<NodeId> local_node_id)
+/// @brief Makes a new CAN transport instance.
+///
+/// NB! Lifetime of the transport instance must never outlive `memory` and `media` instances.
+///
+/// @param memory Reference to a polymorphic memory resource to use for all allocations.
+/// @param media Collection of redundant media interfaces to use.
+/// @param tx_capacity Total number of frames that can be queued for transmission.
```
pavel-kirienko, line 653:
```suggestion
/// @param tx_capacity Total number of frames that can be queued for transmission per `IMedia` instance.
```

include/libcyphal/transport/can/transport.hpp
```diff
@@ -654,14 +644,22 @@ class TransportImpl final : public ICanTransport, private TransportDelegate  //
...
+/// @param local_node_id Optional id of the local node. Could be set (once!) later by `setLocalNodeId` call.
```
pavel-kirienko, line 654:
```suggestion
/// @param local_node_id Optional node-ID of the local node. Could be set (once!) later by `setLocalNodeId` call.
```

It is best to use the terms from the Specification verbatim.

Although, it makes me wonder if we should remove this now that we have `setLocalNodeId()`?

include/libcyphal/transport/transport.hpp
```diff
@@ -61,20 +61,61 @@ class ITransport : public IRunnable
...
-        const ResponseTxParams& params) = 0;
+    virtual cetl::optional<ArgumentError> setLocalNodeId(const NodeId node_id) noexcept = 0;
+
+    /// @brief Makes a message receive (RX) session.
+    ///
+    /// Lifetime of the RX session must never outlive this transport interface.
+    ///
+    /// @param params Various initial parameters for the message RX session (like subject id and extent size).
+    /// @return Either a unique interface pointer to the message RX session, or an error.
+    ///
```
pavel-kirienko, line 72:
```suggestion
```

Do these comments serve any purpose at all? It is evident from the type of the method that it takes params and returns either an interface unique ptr or an error, why spell it out? For better SNR we should only write things that are worth writing.

@thirtytwobits this is an example of an unhelpful doc comment; I suggest we avoid these.

include/libcyphal/transport/can/transport.hpp
serges147: I think it is okey - you might have your node id right from the beginning (from registry or whatever persisting storage).

pavel-kirienko: But in that case you can just call setLocalNodeId later

include/libcyphal/transport/transport.hpp
serges147: Yes, I need them for two purposes:
1. Please see that, in addition to "obvious" stuff, there is also extra note about lifetime of result sessions regarding parent transport.
2. Doxygen - docs generation is just bad without explicit mentioning of public functions.
3. Last but not least - there is less confusion in rules what we document and what we don't - it's much easier to go with "always document public stuff" comparing to fuzzy "it is evident from ..."

pavel-kirienko: My objection related only to the two lines I commented on -- the ones that begin with `@param` and `@return`; the rest is useful and should be kept. I think this addresses points 1 and 2.

serges147: oh, ok, got it - will remove these pairs of lines.

include/libcyphal/transport/can/transport.hpp
serges147: ok, it's done

include/libcyphal/transport/scattered_buffer.hpp
```diff
@@ -167,11 +165,18 @@ class ScatteredBuffer final
     /// @param length_bytes The number of bytes to copy.
     /// @return The number of bytes copied.
     ///
-    CETL_NODISCARD std::size_t copy(const std::size_t offset_bytes,
-                                    void* const       destination,
-                                    const std::size_t length_bytes) const
+    /// NOSONAR cpp:S5008 below currently unavoidable. Could be fixed if nunavut provides `cetl::byte*` support.
```
lydia-at-amazon, line 168: It looks like we're using `cetl::byte*` in other places in this PR now. Does this comment still apply?

lydia-at-amazon:
nit: Title of PR: `No CETL_NODISCARD for the most of public api` --> `No CETL_NODISCARD for most of the public api`

include/libcyphal/transport/scattered_buffer.hpp
serges147: Yes, this comment still applies. Public facing `ScatteredBuffer` still uses `void*` as a second parameter. Internal stuff (aka `detail`) is switched to `cetl::byte*`.

include/libcyphal/transport/transport.hpp
```diff
@@ -61,20 +61,43 @@ class ITransport : public IRunnable
...
-        const RequestTxParams& params) = 0;
-    CETL_NODISCARD virtual Expected<UniquePtr<IResponseRxSession>, AnyError> makeResponseRxSession(
-        const ResponseRxParams& params) = 0;
-    CETL_NODISCARD virtual Expected<UniquePtr<IResponseTxSession>, AnyError> makeResponseTxSession(
-        const ResponseTxParams& params) = 0;
+    virtual cetl::optional<ArgumentError> setLocalNodeId(const NodeId node_id) noexcept = 0;
+
+    /// @brief Makes a message receive (RX) session.
+    ///
+    /// Lifetime of the RX session must never outlive this transport interface.
```
pavel-kirienko, line 68:
```suggestion
    /// The RX session must never outlive this transport interface.
```

pavel-kirienko (approved):

lydia-at-amazon (approved):

PR #361, opened 2024-06-06
Issue 344 & 345: fallible run.

serges147:
Declared additional error-related types and methods.
Also integration with the latest state of CETL to address issue # 351.


include/libcyphal/runnable.hpp
```diff
@@ -6,24 +6,34 @@
...
+#include <cetl/unbounded_variant.hpp>
+
+#include <cstddef>
+
 namespace libcyphal
 {
 
 class IRunnable
 {
+    static constexpr std::size_t MaxErrorSize = sizeof(transport::AnyError);
```
serges147, line 22: Currently this is only one error type here, but my idea is that it will be like following in the future:
```
std::max(sizeof(transport::AnyError), sizeof(XxxLayer::YyyError), ...)
```
when there will be other layers specific error types added.

include/libcyphal/transport/can/msg_rx_session.hpp
```diff
@@ -40,7 +41,7 @@ namespace detail
 /// NOSONAR cpp:S4963 for below `class MessageRxSession` - we do directly handle resources here;
 /// namely: in destructor we have to unsubscribe, as well as let delegate to know this fact.
 ///
-class MessageRxSession final : public IMessageRxSession, private IRxSessionDelegate  // NOSONAR cpp:S4963
+class MessageRxSession final : private IRxSessionDelegate, public IMessageRxSession  // NOSONAR cpp:S4963
```
serges147, line 44: Now, `IRxSessionDelegate` base class finally is the very first in the inheritance list - this becomes possible by switching to `cetl::pmr::InterfacePtr`.

include/libcyphal/transport/can/transport.hpp
```diff
@@ -226,6 +237,11 @@ class TransportImpl final : public ICanTransport, private TransportDelegate  //
                               CANARD_NODE_ID_MAX + 1};
     }
 
+    void setTransientErrorHandler(TransientErrorHandler handler) override
```
serges147, line 240: Proper docs will come in the next PR when there will be more experience. For now this is just blueprint.

include/libcyphal/transport/can/transport.hpp
```diff
@@ -372,6 +395,13 @@ class TransportImpl final : public ICanTransport, private TransportDelegate  //
 
     // MARK: Privates:
 
+    template <typename ErrorVariant>
+    CETL_NODISCARD static AnyError anyErrorFromVariant(ErrorVariant&& other_error_var)
+    {
+        return cetl::visit([](auto&& error) -> AnyError { return std::forward<decltype(error)>(error); },
+                           std::forward<ErrorVariant>(other_error_var));
+    }
```
serges147, line 403: This helper nicely converts subset of error types (like `MediaError` alternatives) to the broader `transport::AnyError` variant.

include/libcyphal/transport/errors.hpp
```diff
@@ -30,10 +34,17 @@ struct MemoryError final
...
-    std::uint32_t code{};
+protected:
+    IPlatformError()                                     = default;
+    ~IPlatformError()                                    = default;
+    IPlatformError(const IPlatformError&)                = default;
+    IPlatformError(IPlatformError&&) noexcept            = default;
+    IPlatformError& operator=(const IPlatformError&)     = default;
+    IPlatformError& operator=(IPlatformError&&) noexcept = default;
 };
+using PlatformError = ImplementationCell<IPlatformError, cetl::unbounded_variant<sizeof(void*) * 3>>;
```
serges147, line 47: This `3` is magic - I guess we will adjust on-fly if some concrete implementation of `IPlatformError` doesn't fit anymore.

include/libcyphal/transport/errors.hpp
```diff
@@ -64,4 +75,17 @@ using MediaError = cetl::variant<ArgumentError, PlatformError, CapacityError>;
...
+namespace cetl
+{
+
+// C6271889-BCF8-43A9-8D79-FA64FC3EFD93
+template <>
+constexpr type_id type_id_getter<libcyphal::transport::AnyError>() noexcept
+{
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
+    return {0xC6, 0x27, 0x18, 0x89, 0xBC, 0xF8, 0x43, 0xA9, 0x8D, 0x79, 0xFA, 0x64, 0xFC, 0x3E, 0xFD, 0x93};
+}
```
serges147, line 87: We put it inside of the `MaybeError` (which is `unbounded_variant`) - hence the type id stuff.

test/unittest/transport/can/test_can_msg_rx_session.cpp
```diff
@@ -67,8 +69,7 @@ class TestCanMsgRxSession : public testing::Test
     void TearDown() override
     {
         EXPECT_THAT(mr_.allocations, IsEmpty());
-        // TODO: Uncomment this when PMR deleter is fixed.
-        // EXPECT_EQ(mr_.total_allocated_bytes, mr_.total_deallocated_bytes);
+        EXPECT_THAT(mr_.total_allocated_bytes, mr_.total_deallocated_bytes);
```
serges147, line 72: `InterfacePtr` deleter now behaves properly. All such todos are resolved.

include/libcyphal/runnable.hpp
```diff
@@ -6,24 +6,34 @@
...
+
 public:
+    using MaybeError = cetl::unbounded_variant<MaxErrorSize>;
```
pavel-kirienko, line 25:
```suggestion
    /// Empty unbounded_variant represents success.
    using MaybeError = cetl::unbounded_variant<MaxErrorSize>;
```

include/libcyphal/transport/can/transport.hpp
pavel-kirienko, line 240:
```suggestion
    void setTransientErrorHandler(const TransientErrorHandler& handler) override
```

include/libcyphal/transport/transport.hpp
```diff
@@ -62,6 +87,8 @@ class ITransport : public IRunnable
     ///
     virtual cetl::optional<ArgumentError> setLocalNodeId(const NodeId node_id) noexcept = 0;
 
+    virtual void setTransientErrorHandler(TransientErrorHandler handler) = 0;
```
pavel-kirienko, line 90:
```suggestion
    virtual void setTransientErrorHandler(const TransientErrorHandler& handler) = 0;
```

The docs are missing, should we expect them later?

include/libcyphal/types.hpp
```diff
@@ -50,12 +51,54 @@ using TimePoint = MonotonicClock::time_point;
...
 template <typename Success, typename Failure>
 using Expected = cetl::variant<Success, Failure>;
 
+/// A generalized implementation of https://www.fluentcpp.com/2021/01/29/inheritance-without-pointers/
+/// that works with any `cetl::unbounded_variant`.
+///
+/// The instance is always initialized with a valid value, but it may turn valueless if the value is moved.
+/// The AnyUbVar type can be a `cetl::unbounded_variant`.
+///
+template <typename Interface, typename AnyUbVar>
```
pavel-kirienko, line 66: Shouldn't we call this any just `Any`? Looking at the usages I think any any would fit, like `std::any`.

include/libcyphal/transport/can/transport.hpp
```diff
@@ -282,11 +298,18 @@ class TransportImpl final : public ICanTransport, private TransportDelegate  //
...
     {
         runMediaTransmit(now);
         runMediaReceive();
-        runMediaFilters();
+
+        cetl::optional<AnyError> any_error = runMediaFilters();
+        if (any_error.has_value())
+        {
+            return any_error.value();
+        }
```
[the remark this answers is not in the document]
pavel-kirienko: I don't get this part though; shouldn't we return `AnyErrorReport` here to match the transient error handler?

pavel-kirienko (commented):
Note for others: for the error handling discussion, please refer here:

- https://github.com/OpenCyphal-Garage/libcyphal/issues/344
- https://github.com/OpenCyphal-Garage/libcyphal/issues/345

include/libcyphal/runnable.hpp
serges147: ok, I'll make it constant (so that current form of `AnyError` will fit).

>if we run over the limit we'll get a build error.
Yes, it will fail to build.

include/libcyphal/transport/can/transport.hpp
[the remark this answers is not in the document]
serges147: I believe that result of generic `run` (which is `MaybeError`) shouldn't have such details - this is transport layer handler what for (with it's knowledge what *network* layer `culprit` could be). Yes, we probably could embed `AnyErrorReport` into `MaybeError`, but I'm very worrying about amount of indirection, and furthermore that higher level code needs to know about lower level "transport" stuff (like CAN specific `transport::can::IMedia*` as a culprit). Here is an example (even without `nullptr` checking):
```
auto maybe_platform_err = cetl::get_if<transport::IPlatformError>(cetl::get_if<transport::AnyErrorReport>(&me)→error));
auto maybe_can_media = get_if<transport::can::IMedia*>IPlatformError>(
    *get_if<transport::AnyError>(get_if<AnyErrorReport>(&me)));
```
and `maybe_can_media` here is actually double pointer like `transport::can::IMedia**`.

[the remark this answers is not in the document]
serges147: Also, notice that I embedded `AnyErrorReport` into `transport::ITransport` - the idea is that this report is only some transport concern (with its expected **internal** knowledge about its possible media or whatever **its level** entity culprit might be).

include/libcyphal/transport/can/transport.hpp
serges147: This was done deliberately like this. Notice that all I do is just store the handler, so in such case "by value and then move" is normally recommended approach.

include/libcyphal/transport/transport.hpp
serges147: >Proper docs will come in the next PR when there will be more experience. For now this is just blueprint.

include/libcyphal/transport/can/transport.hpp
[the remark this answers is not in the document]
pavel-kirienko: image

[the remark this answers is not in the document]
pavel-kirienko: >but probably we can write a visit function for unbounded_variant

If we do that, it would be rather different compared to the variant visitor because we don't know the set of possible types at compile time. We could supply the set of type alternatives at the time of invoking the visitor itself and provide a fallback case when there's no match but it's probably not going to be super easy to use.

pavel-kirienko (approved):

include/libcyphal/transport/can/transport.hpp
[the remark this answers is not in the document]
serges147: Yup, that is exactly what I have in mind.

pavel-kirienko (approved):

include/libcyphal/transport/errors.hpp
```diff
@@ -30,10 +34,17 @@ struct MemoryError final
 struct CapacityError final
 {};
 
-struct PlatformError final
+class IPlatformError
 {
```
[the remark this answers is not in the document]
serges147: I added `std::uint32_ code() const noexcept` for now. Needed for unit tests and coverage. Will refine later if needed.

cmake/modules/Findcetl.cmake
```diff
@@ -6,7 +6,7 @@
 
 include(FetchContent)
 set(cetl_GIT_REPOSITORY "https://github.com/OpenCyphal/cetl.git")
-set(cetl_GIT_TAG "63a118441a80078ebfeabf509c342bc0480a50b2")
+set(cetl_GIT_TAG "5aa0a5b62427547b63cbbf966bcc4cde2abe396e")
```
serges147, line 9: The latest CETL `main` branch commit.

include/libcyphal/transport/errors.hpp
lydia-at-amazon: Is there at least some basis for why 3 was chosen?

include/libcyphal/transport/svc_sessions.hpp
```diff
@@ -94,6 +135,10 @@ class IResponseTxSession : public ITxSession
     ///
     virtual cetl::optional<AnyError> send(const ServiceTransferMetadata& metadata,
```
lydia-at-amazon, line 136: Unrelated to these changes, but who calls `send()`?

lydia-at-amazon (approved):

include/libcyphal/transport/errors.hpp
serges147: Most simple lambdas (capturing f.e. `this`) will fit into 24 bytes.

include/libcyphal/transport/svc_sessions.hpp
serges147: Either user (if she works with so low "transport" level of libcyphal), or upper "presentation" layer (which is the future work after transport layer completion). OR, unit tests of cause ;)

include/libcyphal/transport/errors.hpp
pavel-kirienko: >Is there at least some basis for why 3 was chosen?

vibes-driven design

PR #362, opened 2024-06-10
Address all `TODO: Add error handling`

serges147:
- Extended `AnyErrorReport` with additional `operation` and `media_index` fields.
- More error-handling-related documentation

image


include/libcyphal/transport/can/can_transport.hpp
```diff
@@ -0,0 +1,141 @@
...
+    /// - A reference to the entity that has caused this error.
+    ///
+    struct TransientErrorReport
+    {
+        /// @brief Error report about pushing a message to a TX session.
+        struct CanardTxPush
+        {
+            AnyError        error;
+            std::uint8_t    media_index;
+            CanardInstance& culprit;
```
pavel-kirienko, line 46: Maybe we should make these pointers to ensure the error reports are copyable?

include/libcyphal/transport/can/can_transport.hpp
```diff
@@ -0,0 +1,141 @@
...
+    ICanTransport& operator=(const ICanTransport&)     = delete;
+    ICanTransport& operator=(ICanTransport&&) noexcept = delete;
+
+    /// Sets new transient error handler.
+    ///
+    /// If the handler is set, it will be called by the transport layer when a transient media related error occurs.
+    /// If the handler is not set (default mode), the transport will ignore such errors and continue
+    /// its current process in a "best-effort" manner, namely in assumption that
+    /// either other redundant media (if any) will deliver what is needed,
+    /// or later retry (aka next `run`) of the operation will resolve the issue.
```
pavel-kirienko, line 127: I don't think it's sustainable to ignore errors by default. If the client neglected or forgot to install the handler, the default behavior should be to escalate each error such that it breaks the sending/running/receiving loop. May I recommend we define a default transient error handler and set it in the constructor, such that there is always some handler available, allowing the client to override it if necessary? Having the handler always available could also slightly simplify the internals.

---

I looked at the actual implementation after writing the above comment and noticed that it appears to be implemented correctly: if the handler is not installed, the errors break the loop. Am I misreading the code?

serges147: Yes, I also believe that implementation is correct. I need rewrite this comment to reflect actual implementation.

include/libcyphal/transport/can/can_transport.hpp
serges147: Reports are copy constructible, but yes they are not copy **assign**able. IMHO it's ok limitation, especially b/c I didn't want to expose raw pointer interface (of `CanardInstance` or `IMedia`) - that is why I made them as `&` in the first place.

pavel-kirienko (approved):

lydia-at-amazon (approved):

PR #363, opened 2024-06-21
Implementation of UDP TX

serges147:
- UDP implementation:
  - `TransportDelegate` & `TransportImp`.
  - Defined `udp::IMedia`, `udp::ITxSocket` & `upd::IRxSocket` protocols.
  - Implementation of all 3 TX sessions.
- Fixed various Sonar issues (after switching to correct "libcyphal" quality profile at Sonar cloud).
- Increased CAN-related code coverage to 100% (files, lines, functions & branches!).
- A bit less of code duplication (f.e. `libcyphal::detail::UniquePtrSpec` helper).
- Simplifying "swapping" behavior by using `std::exchange`.

Here is code coverage. Note that low (red) coverage at RX is currently expected - will be addressed by near future next PRs.
image


include/libcyphal/transport/udp/udp_transport_impl.hpp
```diff
@@ -0,0 +1,708 @@
...
+        while (const UdpardTxItem* const maybe_item = ::udpardTxPeek(&udpard_tx))
+        {
+            UdpardTxItem* const item = ::udpardTxPop(&udpard_tx, maybe_item);
+            ::udpardTxFree(memoryResources().fragment, item);
+        }
+    }
+
+    /// @brief Runs transmission loop for each redundant media interface.
+    ///
+    CETL_NODISCARD cetl::optional<AnyError> runMediaTransmit(const TimePoint now)
```
thirtytwobits, line 570: Oh dear. We've missed something important in `IRunnable`. We don't have a timeout. This means we cannot schedule this work on single-threaded systems nor can we implement variable schedulers in the future per https://forum.opencyphal.org/t/design-review-execution-model-for-libcyphal/1918 . Even if we do add a timeout we haven't defined how to get a clock for this implementation yet, or have we?

include/libcyphal/transport/can/can_transport.hpp
```diff
@@ -6,9 +6,9 @@
 #ifndef LIBCYPHAL_TRANSPORT_CAN_TRANSPORT_HPP_INCLUDED
 #define LIBCYPHAL_TRANSPORT_CAN_TRANSPORT_HPP_INCLUDED
 
-#include "libcyphal/transport/can/media.hpp"
 #include "libcyphal/transport/errors.hpp"
 #include "libcyphal/transport/transport.hpp"
+#include "media.hpp"
```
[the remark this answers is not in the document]
serges147: B/c this "island" of `"..."` includes are **relative** ones:
- either relative from **current** file location (this is why just `"media.hpp"` - it is a sibling)
- or from root of includes (aka `<repo>/include/`)

Note that instead of f.e. `#include "libcyphal/transport/errors.hpp"` another option will be `#include "../errors.hpp"`, but I decided against usage of such `..` (grand-)parent folder references. But in case of sibling/children includes IMO it makes sense use shorter version - f.e. if I would need rename ".../can/..." part of the path then it will zero changes for those files inside of the "can" folder.

include/libcyphal/transport/udp/udp_transport_impl.hpp
serges147: Scott, is it (the timeout) is really needed? I thought that we have requirement that no entity inside of a `run` is allowed to block, so if you run a runnable frequently enough there will be not too much to do on each such run. With this approach I don't see why it can't be run on a single-threaded systems (aka super-loop).

Also, I believe that introducing such timeout (for a single run, if I guessed correctly what you meant) will mean that you need to bring/pass it almost everywhere inside (essentially to any method/entity which might do any loop OR anything at all which might overrun allowed slot of time). I believe it will pollute codebase a lot, and will make it harder unit test (b/c you need to cover any potential break condition due to timeout). Error handling probably will be affected as well...

Here is another idea - what if (user-provided) `IMedia` implementation could (if needed) also monitor current allowed time-slot for a run, and if it "sees" (on its `push`/`send`/`pop` etc invocations) that the time-slot is or close enough to be exhausted than it will start quickly return "not ready" - this will quickly unwind/break whatever loops we have in libcyphal transport implementation. Also such user-provided `IMedia` implementation probably could have sense of its timing, so below concern won't be libcyphals one:
>_we haven't defined how to get a clock for this implementation yet_

serges147: Anyway, I believe Scott this discussion is beyond this PR.

pavel-kirienko: @thirtytwobits please check this and the following section titled "Second proposal: Stateful": https://github.com/OpenCyphal-Garage/libcyphal/blob/main/docs/design/readme.md#execution-flow-and-the-runner-interface

>This approach is not compatible with the one-thread-per-runnable execution strategy because all entities that share a given `IMultiplexer` will necessarily have to share the same executor. It will still be possible to run one transport instance per thread, though.

A transport (or some presentation-layer entity) can run in a separate thread similarly to what you described in that thread; where this solution differs is that we never block inside LibCyphal (it's a hard policy mentioned in the introductory part of the document -- everything is nonblocking) but instead we block in the platform-specific code implementing `IMultiplexer`. The client code will block on its multiplexer implementation until there is an event to handle or the next polling deadline is reached, then `IRunnable::run` will be invoked once to handle the events, then the execution thread is blocked again until the next event.

A POSIX-specific implementation of `IMultiplexer` will be probably based on `poll`; such an implementation can provide a utility method that invokes `run` in a loop continulusly, timed by the `poll` call.

thirtytwobits: We can move this discussion to another forum but my concern is that the bounds of these while-loops is affected by the amount of I/O which couples the WCET of each IRunnable to receive rates (coupling between vehicle components) and publish rates (coupling between software components). How do we ensure that these loops have a bounded execution time without using a time bound?

include/libcyphal/transport/can/can_transport.hpp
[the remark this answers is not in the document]
thirtytwobits: I do not believe that the typical behaviour of looking first in the current folder is actually controlled by the c++ standard.

include/libcyphal/transport/udp/udp_transport_impl.hpp
thirtytwobits: > Also, I believe that introducing such timeout (for a single run, if I guessed correctly what you meant) will mean that you need to bring/pass it almost everywhere inside (essentially to any method/entity which might do any loop OR anything at all which might overrun allowed slot of time).

Not necessarily. What we need to ensure is that any loop which can change in size based on external inputs can be bounded, especially if the bounds can change while we are in the loop. In this example, `::udpardTxPeek(&media.udpard_tx())` is bounding the loop which – theoretically if not practically – may continue to grow indefinitely unless we're enforcing single-thread access to each udpard instance?

thirtytwobits: > Here is another idea - what if (user-provided) IMedia implementation could (if needed) also monitor current allowed time-slot for a run, and if it "sees" (on its push/send/pop etc invocations) that the time-slot is or close enough to be exhausted then it will start quickly return "not ready" or "nullopt" data 

Hmm. How would the IMedia know that a new run-cycle had started?

include/libcyphal/transport/udp/udp_transport_impl.hpp
```diff
@@ -0,0 +1,708 @@
...
+        {
+            return MemoryError{};
+        }
+
+        cetl::optional<AnyError> opt_any_error;
+
+        for (Media& some_media : media_array_)
+        {
+            opt_any_error = withEnsureMediaTxSocket(some_media,
+                                                    [this, &tx_metadata_var, &payload](auto& media, auto& tx_socket)
```
thirtytwobits, line 373: This lambda is probably going to exceed most small-function optimizations and end up calling malloc. Can we make this a function on the callstack that this lambda delegates to? Nobody is storing this lambda after the callstack unwinds, right?

pavel-kirienko: This code does not make use of any type-erasing function containers, so the matters of small-size optimizations are not applicable here. There is nothing that could use dynamic memory here.

include/libcyphal/transport/udp/udp_transport_impl.hpp
pavel-kirienko: >We can move this discussion to another forum but my concern is that the bounds of these while-loops is affected by the amount of I/O which couples the WCET of each IRunnable to receive rates (coupling between vehicle components) and publish rates (coupling between software components). How do we ensure that these loops have a bounded execution time without using a time bound?

Ah, I understand now what we're after here. Per the design proposed in the design doc, once the multiplexer has signaled an event, we will keep `run`ning the `IRunnable`s related to that multiplexer until all work is done. Indeed, there is no way to communicate a deadline --- the call will not return until the RX queue in the media is emptied and there are no pending TX frames left OR the media TX queue is full.

We could retrofit a deadline here quite easily, but think about the temporal granularity here. Our runnables do the job in small chunks --- we need to either push or pop a frame per iteration. Is it acceptable to honor the deadline with the granularity that is defined by the time it takes to process one work item (one frame, etc)?

At the transport layer we have no activities that could take an undefined amount of time to complete (the time complexity of each operation is well-defined in the lizards), but then on the presentation layer we may be dealing with more complex entities such as user callbacks and so on. Is it still acceptable to provide the deadline guarantees whose resolution is essentially dependent on the user code?

thirtytwobits: >  Is it acceptable to honor the deadline with the granularity that is defined by the time it takes to process one work item (one frame, etc)?

Yes. Absolutely. This is the dominant term so as long as we bound the WCET to ±1 unit of work we'll meet our objectives of bounded-realtime if not hard-realtime.

include/libcyphal/transport/can/can_transport.hpp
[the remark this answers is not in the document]
pavel-kirienko: @serges147 This is true. The standard is incredibly vague on the specifics of the search behavior.

image

include/libcyphal/transport/udp/udp_transport_impl.hpp
thirtytwobits: It's unfair of me to ask this without an ability to test so I'll retract this comment for now but we may have to revisit this in the future.

We have seen that there is a threshold of complexity or perhaps some way of referencing the lambda that causes an `std::function` to be generated instead. `std::function` does have a small-function optimization that will fall-back to malloc.

Found this with some googling: https://stackoverflow.com/questions/70590194/small-object-optimization-useless-in-using-stdfunction

pavel-kirienko: I would like to learn more about this. The referenced StackOverflow post does not seem to be related (unless I'm misreading it) because it takes about std::function specifically, which we do not use. Are you saying a plain lambda can be boxed into an std::function? That sounds wild.

include/libcyphal/transport/udp/udp_transport_impl.hpp
pavel-kirienko: Moved into #364

include/libcyphal/transport/udp/udp_transport_impl.hpp
thirtytwobits: It's that lambdas can only be converted into function pointers if they do not capture. Once you capture the closure will have to be stored somewhere. This means that, when passing a lambda as an argument, any overrides that take `std::function` will be selected which may or may not be able to store the capture in the implementation's small-object buffer.

pavel-kirienko: I understand that; it's the documented behavior of std::function. We are not using std::function so it would seem that this is a non-issue, or am I missing something?

pavel-kirienko: We forward the lambda directly without any type-erasure.

include/libcyphal/transport/can/can_transport.hpp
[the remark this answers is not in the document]
serges147: ok, but I believe that a lot of existing c++ codebases will stop to compile if `xxx.hpp` won't be searched for `#include "xxx.hpp"` at sibling `xxx.cpp` anymore...

Also, here is what I used as a guideline: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-incform
Although, if I followed this guideline strict then I would use more `<>`... Instead... but I think found middle ground, namely:
- use `""` syntax for your own "local" stuff
- use `<>` for anything which is external, like: CETL, lizards, std, or **even** libcyphal includes themselves if they are referenced in unit tests (I treat libcyphal's unit tests as "external" client/user of the headers).

Also please note everywhere at libcyphal (and it's unit tests) this approach is applied consistently, so it's not like here I use one approach and in other place different one - I indeed care about inclusions of headers very much, grouping them to separate logical islands, keeping their order in very strict specific order, and eliminating anything that became unused.

Anyway, please let me know if you guys decide that, still, we should not rely on "search first local/relative" behavior, but instead ALWAYS use "fullpath" `#include "libcyphal/xxx/yyy.hpp"` regardless from where it was included at a `libcyphal/...` source file). I will apply (consistently of course) such changes everywhere. Thanks!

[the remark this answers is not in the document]
pavel-kirienko: I wasn't aware of that core guidelines rule. The guidelines are not the standard but as they are published by Stroustrup and Sutter, we can probably rely on them here?

include/libcyphal/transport/udp/tx_rx_sockets.hpp
```diff
@@ -0,0 +1,109 @@
...
+namespace udp
+{
+
+/// @brief Defines IP endpoint information in use for multicast transmissions.
+///
+struct IpEndpoint final
+{
+    std::uint32_t ip_address;
+    std::uint16_t udp_port;
+};
```
pavel-kirienko, line 32: I think it would be wiser to alias `UdpardUDPIPEndpoint` so that you don't have to convert there and back in your code. Also I think it's more accurate to call it a UDP/IP endpoint, not just an IP endpoint.

pavel-kirienko (approved):

include/libcyphal/transport/udp/tx_rx_sockets.hpp
serges147: My thinking was that we are under `udp::` namespace, so the type is kinda already has UDP, like `upd::IpEndpoint`...

include/libcyphal/transport/udp/delegate.hpp
```diff
@@ -0,0 +1,259 @@
...
+        UdpardTransferID  transfer_id;
+    };
+    struct Respond
+    {
+        UdpardMicrosecond deadline_us;
+        UdpardPriority    priority;
+        UdpardPortID      service_id;
+        UdpardNodeID      client_node_id;
+        UdpardTransferID  transfer_id;
+    };
```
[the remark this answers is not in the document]
serges147: Am I right that you are proposing to extract `deadline_us`, `priority` & `transfer_id` to some `BaseMetadata`?

In general I saw and was aware of this duplication, but decided that a bit of duplication in `detail` won't hurt but instead will give clear independence/separation of the cases (which are instantiated in 3 different places, and in the end consumed by 3 different `operator()`-s)...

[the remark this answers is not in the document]
serges147: @pavel-kirienko ?

serges147:
@thirtytwobits Scott, how is PR overall? Aproveable? ;)

thirtytwobits (approved):

PR #366, opened 2024-07-01
General structure for UDP RX sessions

serges147:
RX sessions functionality is not complete in this PR, but general structure is there. Msg RX session is implemented a bit further (but still not yet integrated with media sockets - will be the next PR).

- Added new implementation of a `ScatteredBuffer::IStorage` protocol - the `UdpardMemory` class manages (owns) scattered fragments received from libudpard.
- Added CAVL tree implementation - in use to ensure a single RX session per its kind & port id. New `detail::SessionsTree` wraps the tree by exposing only sessions relevant api, as well as managing tree node memory.

Also:
- minor fixes based on previous PR review


include/libcyphal/transport/udp/msg_rx_session.hpp
```diff
@@ -122,9 +123,20 @@ class MessageRxSession final : private IRxSessionDelegate, public IMessageRxSess
 
     // MARK: IRxSessionDelegate
 
-    void acceptRxTransfer(const UdpardRxTransfer&) override
+    void acceptRxTransfer(UdpardRxTransfer& inout_transfer) override
```
serges147, line 126: This is non-`const` now deliberately - we move memory fragments from this RX transfer into the `UdpardMemory` instance mentioned previously. Hence parameter called as `inout_transfer`.

include/libcyphal/transport/udp/tx_rx_sockets.hpp
```diff
@@ -96,6 +101,27 @@ class IRxSocket
...
+    /// @brief Defines the error types that may occur during the `receive` operation.
+    ///
+    using ReceiveFailure = cetl::variant<PlatformError, ArgumentError, MemoryError>;
+
+    struct ReceiveSuccess
+    {
+        TimePoint   timestamp;
+        std::size_t payload_size;
+        // TODO: Use CETL pmr instead of std::unique_ptr.
+        std::unique_ptr<cetl::byte[]> payload;  // NOLINT(*-avoid-c-arrays)
```
serges147, line 113: This is very temporary solution (and of course it's heap based - hence the `// TODO`). I need figure out how better support array allocations at our CETL pmr (it seems that currently it's not yet supported by the CETL).

include/libcyphal/transport/udp/tx_rx_sockets.hpp
```diff
@@ -96,6 +101,27 @@ class IRxSocket
     IRxSocket& operator=(const IRxSocket&)     = delete;
     IRxSocket& operator=(IRxSocket&&) noexcept = delete;
 
+    /// @brief Defines the error types that may occur during the `receive` operation.
+    ///
+    using ReceiveFailure = cetl::variant<PlatformError, ArgumentError, MemoryError>;
```
serges147, line 106: If you don't mind I would like to start renaming several `XxxError`-s to similar but with `...Failure` suffix when it comes to a variant of errors. Idea is that it will be more clear when...
- it's a "primitive" error - use current approach with `Error` suffix
- it's a higher level variant of the above errors - use `Failure` suffix.

This is the first bird, and if we agreed on what I'm proposing then I will rename also `can::MediaError` → `can::MediaFailure`, `transport::AnyError` → `transport::AnyFailure` and couple more.

include/libcyphal/transport/udp/udp_transport_impl.hpp
```diff
@@ -175,6 +176,9 @@ class TransportImpl final : private TransportDelegate, public IUdpTransport  //
                   MediaArray&&           media_array)
         : TransportDelegate{memory_resources}
         , media_array_{std::move(media_array)}
+        , msg_rx_session_nodes_{memory_resources.general}
+        , svc_request_rx_session_nodes_{memory_resources.general}
+        , svc_response_rx_session_nodes_{memory_resources.general}
```
serges147, line 181: Currently, it might seems that we can go with just one combined tree, but I anticipate that their node types might deviate from each other (although currently they are exactly the same - all derived from `BaseNode`; without their own specifics). I'll keep in mind these 3 trees and if needed will make proper changes.

include/libcyphal/transport/udp/delegate.hpp
```diff
@@ -69,7 +75,150 @@ struct AnyUdpardTxMetadata
...
+        {
+            struct Destroyed
+            {
+                PortId service_id;
+            };
+
+            using Variant = cetl::variant<Destroyed>;
+        };
+
+        using Variant = cetl::variant<Message::Variant, Request::Variant, Response::Variant>;
```
serges147, line 218: Here we have variant of variant. My idea is to keep session events kinda grouped...
- first, by their "kind" type (message, svc request & response)
- second, by their action type (inside of its kind)

Rationale:
- kinda more structured (hierarchical) way IMHO than "flat" approach
- extending actions of one kind (I anticipate that Req/Res probably will have more actions soon) won't touch other kind
- 2 `std::visit`-s (see below in code) quite nicely "flattens" the hierarchy when it comes to an event handler functor.

include/libcyphal/transport/udp/delegate.hpp
```diff
@@ -69,7 +75,150 @@ struct AnyUdpardTxMetadata
...
+            };
+
+            using Variant = cetl::variant<Destroyed>;
+        };
+
+        struct Response
+        {
+            struct Destroyed
+            {
+                PortId service_id;
```
serges147, line 212: They all look the same, but...
- we mostly just need different types (for the visit pattern)
- code duplication is minimal, and making a base looks like unnecessary (and strictly speaking will require switch to `class`-s
- also see next comment about hierarchy intention

cmake/CMakeCommon.cmake
```diff
@@ -100,6 +100,7 @@ cmake_path(APPEND LIBCYPHAL_ROOT "external" OUTPUT_VARIABLE CETLVAST_EXTERNAL_RO
 if (CETL_ENABLE_DEBUG_ASSERT)
     add_compile_definitions("CETL_ENABLE_DEBUG_ASSERT=1")
 else()
+    add_compile_definitions("CAVL_NO_ASSERT=1")
```
pavel-kirienko, line 103: We should be mindful that in many cases the library will be used without its CMake recipes by merely adding it to the include path list. That bypassess the logic implemented here that is aware that `CETL_ENABLE_DEBUG_ASSERT` implies no `CAVL_NO_ASSERT` and vice versa. Since we're vendoring `cavl.hpp`, should we not modify it to rely on `CETL_ENABLE_DEBUG_ASSERT` instead of `CAVL_NO_ASSERT`?

include/libcyphal/transport/udp/tx_rx_sockets.hpp
pavel-kirienko: So the proposed taxonomy is:

- Result:
  - Success
  - Failure
    - Error X
    - Error Y
    - ...

Maybe it should be documented

include/libcyphal/transport/udp/udp_transport_impl.hpp
pavel-kirienko: As we discussed earlier today, we should expect them to be different indeed; plus, having separate trees reduces their depth which improves WCET

include/libcyphal/transport/udp/tx_rx_sockets.hpp
serges147: Yes, something like that. I'll add todo about documenting the taxonomy, especially if it will fit well.

cmake/CMakeCommon.cmake
serges147: This line is just to disable CAVL asserts for "Coverage" build flavor only.

>should we not modify it to rely on CETL_ENABLE_DEBUG_ASSERT instead of CAVL_NO_ASSERT?
Could you please clarify, are you suggesting to find/replace `CAVL_NO_ASSERT` with `CETL_DEBUG_ASSERT` at `cavl.hpp`? Thnx

include/libcyphal/transport/udp/udp_transport_impl.hpp
```diff
@@ -388,6 +375,16 @@ class TransportImpl final : private TransportDelegate, public IUdpTransport  //
         return opt_any_error;
     }
 
+    void onSessionEvent(const SessionEvent::Variant& event_var) override
+    {
+        cetl::visit(
+            [this](const auto& event) {
+                SessionEventHandler handler_with{*this};
+                cetl::visit(handler_with, event);
```
pavel-kirienko, line 383:
```suggestion
                cetl::visit(cetl::pf17::make_overloaded(
                                [this](const SessionEvent::Message::Destroyed& event) {
                                    msg_rx_session_nodes_.removeNodeFor(event.subject_id);  //
                                },
                                [this](const SessionEvent::Request::Destroyed& event) {
                                    svc_request_rx_session_nodes_.removeNodeFor(event.service_id);
                                },
                                [this](const SessionEvent::Response::Destroyed& event) {
                                    svc_response_rx_session_nodes_.removeNodeFor(event.service_id);
                                }),
                            event);
```

We should have used the same for `TxTransferHandler`, I overlooked this part in the previous PR.

include/libcyphal/transport/udp/sessions_tree.hpp
```diff
@@ -0,0 +1,197 @@
...
+///
+namespace detail
+{
+
+/// @brief Defines a tree of sessions for the UDP transport.
+///
+/// No Sonar cpp:S4963 b/c we do directly handle resources here.
+///
+template <typename Node>
+class SessionsTree final  // NOSONAR cpp:S4963
```
pavel-kirienko, line 38: I'm pretty sure this should be `SessionTree` and the file would be `session_tree.hpp`.

include/libcyphal/transport/udp/delegate.hpp
```diff
@@ -69,7 +75,150 @@ struct AnyUdpardTxMetadata
...
+            return payload_size_;
+        }
+
+        CETL_NODISCARD std::size_t copy(const std::size_t offset_bytes,
+                                        cetl::byte* const destination,
+                                        const std::size_t length_bytes) const override
+        {
+            using FragSpan = const cetl::span<const cetl::byte>;
+
+            // TODO: Use `udpardGather` function when it will be available with offset support.
```
pavel-kirienko, line 127:
```suggestion
            // TODO: Use `udpardGather` function when it will be available with offset support.
            // See https://github.com/OpenCyphal/libudpard/issues/57
```

pavel-kirienko (approved):

lydia-at-amazon (commented):
+1 to Pavel's comment to change `SessionsTree` to `SessionTree`. No additional comments from me at this time. Looks good.

include/libcyphal/transport/udp/sessions_tree.hpp
pavel-kirienko: Actually, should we not promote this as a more general-purpose container? Looking at the implementation I don't see anything session-specific here. We will need a similar container at the presentation layer as well soon...

serges147: There are things which are quite specific:
1. `<Node>` typename expectation:
  a. having `using NodeRef = typename Node::ReferenceWrapper;`
  b. having `.compareWith` by port id
  c. constructable by single port id argument
3. `ensureNewNodeFor` operates with transport level `AnyError` failure

I propose to do now only `SessionsTree` → `SessionTree` renaming. When we will be at presentation, and there will be similar use case for the tree - then try refactor it somehow to be common.

include/libcyphal/transport/udp/udp_transport_impl.hpp
serges147: I agree about `SessionEventHandler` but applying the same for `TxTransferHandler` makes things quite ugly, like this:
image

serges147: Also, idea of these `XxxHandler`-s was to encapsulate handling business logic into small but separate "handler" private sub-entity. So, I'm not quite convinced that using `make_overloaded` is better here...

lydia-at-amazon (approved):

PR #371, opened 2024-08-02
Callbacks by value (no PMR) #verification #docs #sonar

serges147:


include/libcyphal/platform/single_threaded_executor.hpp
```diff
@@ -113,114 +98,176 @@ class SingleThreadedExecutor : public IExecutor  // NOSONAR cpp:S4963
...
+
+            return *this;
+        }
+
+        CallbackNode(const CallbackNode&)            = delete;
+        CallbackNode& operator=(const CallbackNode&) = delete;
+
+        explicit operator Callback::Handle() const noexcept
+        {
+            // Next nolint & NOSONAR are unavoidable: this is opeque handle coversion code.
```
thirtytwobits, line 208: opaque/conversion (sp.)

include/libcyphal/platform/single_threaded_executor.hpp
```diff
@@ -113,114 +98,176 @@ class SingleThreadedExecutor : public IExecutor  // NOSONAR cpp:S4963
...
+            CETL_DEBUG_ASSERT(handle, "");
+
+            // Next nolint & NOSONAR are unavoidable: this is opeque handle coversion code.
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast, performance-no-int-to-ptr)
+            return reinterpret_cast<CallbackNode*>(handle);  // NOSONAR cpp:S3630
+        }
+
+        SingleThreadedExecutor* executor() const noexcept
+        {
+            return executor_;
```
thirtytwobits, line 223: can this be null? If not can it be a reference?

include/libcyphal/executor.hpp
```diff
@@ -204,6 +89,27 @@ class IExecutor : public cetl::rtti_helper<IExecutorTypeIdType>
...
+        /// It is supposed to be used to hide callback implementation,
+        /// and to be passed to the executor for scheduling.
+        /// It also manages lifetime and ownership of the callback registration, so it's movable but not copyable.
+        ///
+        using Any = cetl::unbounded_variant<MaxSize, false /* Copyable */, true /* Movable */>;
+
+        /// @brief Defines opaque handle to a registered callback.
+        ///
+        /// It is supposed to be used internaly (by derived executors)
+        /// to identify/reference the callback in the executor.
```
thirtytwobits, line 109: sp. internally

thirtytwobits (changes requested):
Only thing of import in my review is the raw pointer in single_threaded_executor.

include/libcyphal/executor.hpp
```diff
@@ -219,64 +125,30 @@ class IExecutor : public cetl::rtti_helper<IExecutorTypeIdType>
...
-    ///         or there are not enough resources to schedule the callback.
+    /// @param callback The callback instance (registered at this executor) to be scheduled.
+    ///                 There will be no scheduling if callback is not registered yet
+    ///                 (in its initial default value), or has been reset already.
+    /// @param schedule Contains specifics of how exactly callback will be scheduled (like once, repeatedly etc.).
     ///
-    virtual bool scheduleCallbackById(const Callback::Id                 callback_id,
-                                      const TimePoint                    exec_time,
-                                      const Callback::Schedule::Variant& schedule) = 0;
+    virtual void scheduleCallback(Callback::Any& callback, const Callback::Schedule::Variant& schedule) = 0;
```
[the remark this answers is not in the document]
serges147: Plz see above docs for `callback` parameter - so, initial/default (or `.reset()`-ed) callback instance is the only way how it could kinda "fail". If you think it deserves `bool` (instead of `void`) let me know - initially I did it like this (with `bool`) but later I inclined to just document expectation.

For the "past" question please see documentation for Schedule. In short - it will be executed asap.

include/libcyphal/platform/single_threaded_executor.hpp
serges147: It can't be null, and it can't be (unfortunately) a reference - otherwise it will be impossible to have move assignment operator. But I believe I can make it as a reference in constructor @ line 160 (currently it is a pointer with assert there) - just to emphasize that it can't null.

serges147: Actually, it's possible - see last commit with `std::reference_wrapper`.

include/libcyphal/executor.hpp
```diff
@@ -29,164 +29,49 @@ using IExecutorTypeIdType = cetl::
...
-        ///
-        /// 64 bits should be enough to ensure uniqueness of the callback identifier
-        /// by applying some simple rules (e.g. incrementing counter).
-        ///
-        using Id = std::uint64_t;
-
+    public:
         /// @brief Defines possible variants of callback schedules.
         ///
         struct Schedule
```
pavel-kirienko, line 39: It's a nit but maybe it doesn't need to be nested inside `Callback`? Fully qualified names look unwieldy with deep nesting.

include/libcyphal/executor.hpp
```diff
@@ -29,164 +29,49 @@ using IExecutorTypeIdType = cetl::
...
         {
+            /// @brief Defines schedule which will NOT execute callback function.
+            ///
+            /// Useful as a default value for a non-engaged callback schedule.
+            /// Also can be used to cancel previously scheduled callback.
+            ///
+            struct None
+            {};
```
pavel-kirienko, line 47: This special case doesn't look convincing. I understand that there is some value in being able to unschedule a callback but as far as I understand by now we haven't identified an explicit need for that, and it would be easy to retrofit -- if needed -- later. The troubling thing here is that an extension of any well-structured set of entities with a special case often tends to break the design in unexpected ways at some point.

>*But I couldn’t resist the temptation to put in a null reference, simply because it was so easy to implement.*

include/libcyphal/executor.hpp
```diff
@@ -204,6 +89,27 @@ class IExecutor : public cetl::rtti_helper<IExecutorTypeIdType>
         ///
         using Function = cetl::pmr::function<void(const TimePoint now_time), FunctionMaxSize>;
```
pavel-kirienko, line 90: It is often convenient, especially for periodic callbacks, to know the scheduled time along with the current time. Would it be difficult to replace this with a struct of two fields: the current time and the scheduled time?

include/libcyphal/executor.hpp
```diff
@@ -204,6 +89,27 @@ class IExecutor : public cetl::rtti_helper<IExecutorTypeIdType>
...
+        /// It also manages lifetime and ownership of the callback registration, so it's movable but not copyable.
+        ///
+        using Any = cetl::unbounded_variant<MaxSize, false /* Copyable */, true /* Movable */>;
+
+        /// @brief Defines opaque handle to a registered callback.
+        ///
+        /// It is supposed to be used internally (by derived executors)
+        /// to identify/reference the callback in the executor.
+        ///
+        using Handle = std::uintptr_t;
```
pavel-kirienko, line 111: Casting pointer-to-int and back looks sus. Obviously there's no casting going on here but the `uintptr` named "Handle" obviously invites one to do that. Why do we need this in the interface considering that it's not used by the interface's clients nor the interface itself? If an implementation needs to do `reinterpret_cast<>` (which is not good in itself but that is beside the point), let it define its own Handle type.

Keeping it here contaminates the interface with implementation details, I suggest removing it.

include/libcyphal/platform/single_threaded_executor.hpp
```diff
@@ -113,114 +99,175 @@ class SingleThreadedExecutor : public IExecutor  // NOSONAR cpp:S4963
...
-        removeIfScheduledNode(callback_node);
-        callback_node.schedule() = schedule;
-        insertScheduledNode(callback_node, exec_time);
-
-        return true;
     }
 
-    void removeCallbackById(const Callback::Id callback_id) override
+protected:
+    static Callback::Handle callbackToHandle(Callback::Any& callback) noexcept
```
pavel-kirienko, line 123: Maybe I'm missing something obvious but why can't we just use `CallbackNode*` instead? This pointer casting is unsafe.

include/libcyphal/platform/single_threaded_executor.hpp
```diff
@@ -113,114 +99,175 @@ class SingleThreadedExecutor : public IExecutor  // NOSONAR cpp:S4963
...
+                            },
+                            [this](const Callback::Schedule::Once&) {
+                                //
+                                next_exec_time_ = TimePoint::max();
+                            },
+                            [this, exec_time](const Callback::Schedule::Repeat& repeat) {
+                                //
+                                next_exec_time_ = exec_time + repeat.period;
+                            }),
+                        schedule_);
```
pavel-kirienko, line 265: This is just a silly nit, feel free to ignore, but

```suggestion            next_exec_time_ =
                cetl::visit(cetl::make_overloaded([](const Callback::Schedule::None&) { return TimePoint::max(); },
                                                  [](const Callback::Schedule::Once&) { return TimePoint::max(); },
                                                  [exec_time](const Callback::Schedule::Repeat& repeat) {
                                                      return exec_time + repeat.period;
                                                  }),
                            schedule_);
```

include/libcyphal/executor.hpp
serges147: In the next pr I'll refactor this in favor of optional schedule.

pavel-kirienko: https://github.com/OpenCyphal-Garage/libcyphal/issues/372

include/libcyphal/executor.hpp
serges147: OK, in the next pr I'll add new `Callback::Arg` with two fields.

include/libcyphal/platform/single_threaded_executor.hpp
pavel-kirienko: Moved into https://github.com/OpenCyphal-Garage/libcyphal/issues/373

pavel-kirienko (approved):

include/libcyphal/executor.hpp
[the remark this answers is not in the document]
thirtytwobits: Here's my concern:
```
MyObject* p = new MyObject();
e->scheduleCallback([p]{ delete p; }); <-- if this isn't scheduled then we have a memory leak.
```

thirtytwobits (changes requested):

include/libcyphal/executor.hpp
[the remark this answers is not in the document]
serges147: Should you in the first place use unique ptr for `MyObject`, and capture it by move? Also, even if you would have scheduling `true` result but later call `callback.reset()` (before the callback had a chance to be executed) then you will have the same "leak" (but not with smart ptr).
But in general I've got your concern. ok, I'll try with `bool` return and see how it will go inside libcyphal... but I anticipate complications and surplus of extra corner cases and failure points in originally quite safe places.

Honestly, I would rather have extra bunch of asserts which will fail if you try schedule already reset callback. And also make `.schedule` method as part of `Callback::Any` type (like `callback.shedule(...)` instead of current `executor.scheduleCallback(callback, ...)`) to eliminate what I've describe in docs as "_...must be used only with the same executor..._" - I discussed this move with @pavel-kirienko , and we agreed to do it (but in separate PR).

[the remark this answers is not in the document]
thirtytwobits: Let's open an issue up on this and then proceed with this CR as-is. I'm concerned about the case where the scheduled time can become in the past between the time it is calculated and the time the schedule method is invoked.

thirtytwobits (approved):

Issue #367, opened 2024-07-04
Refactor the execution logic


pavel-kirienko:
The multiplexer as proposed has certain design issues that became apparent only now. The design needs to be slightly adjusted. The updated design will follow Scott's proposed [execution model](https://forum.opencyphal.org/t/design-review-execution-model-for-libcyphal/1918) more closely.

The `IMultiplexer` will be modified to encapsulate not only IO polling but rather scheduling of executable entities in response to 1. input availability; 2. output readiness; 3. delay expiration. As such, the interface will be renamed into `IScheduler` (alternative names: `IExecutor`, `IExecutionPolicy`). The implementation of the interface will be informed by the specifics of the platform (e.g., a POSIX platform may use `poll`, FreeRTOS may use `xQueueCreateSet`/`FreeRTOS_select`, etc.) and by the preferred threading model (single-thread superloop, dedicated single thread for event handling, or a thread pool for the larger systems). Similar to the original LibUAVCAN design, LibCyphal will probably come with a small set of predefined execution policies for common usage scenarios, while allowing the user to easily create their own (perhaps using one of the predefined policies as a starting point).

The `IScheduler` (aka execution policy interface) will allow the library to register a `IRunnable` entity along with a condition for its execution (e.g., a socket became readable or a timeout expired). For now, to simplify the implementation of policies, we introduce a restriction that a single runnable entity can only be registered with either an IO event or a timeout (e.g., cannot schedule execution both on data availability AND a timeout, you have to choose one); later this restriction may be lifted if needed.

The first approximation of the interface is as follows:

```c++
class IScheduler
{
public:
    /// Invoke the specified runnable when the pollable becomes readable/writable.
    /// The future must be kept alive until the event is no longer needed.
    virtual expected<Future, Error> source(IDataSource& io, IRunnable& handler) = 0;
    virtual expected<Future, Error> sink(IDataSink& io, IRunnable& handler) = 0;

    /// Invoke the specified runnable at the specified period or after the delay expiration.
    /// The scheduler shall avoid accumulation of the phase error.
    /// The future must be kept alive until the event is no longer needed.
    virtual expected<Future, Error> repeat(const Duration period, const IRunnable& handler) = 0;
    virtual expected<Future, Error> defer(const Duration delay, const IRunnable& handler) = 0;

    // rule of 5...
};
```

[...]

Issue #365, opened 2024-06-25
Reduce duplication around `TransferMetadata`


pavel-kirienko:
Here:

https://github.com/OpenCyphal-Garage/libcyphal/blob/5e652de5cb6c9172b833fb8426ba619ce99c4963/include/libcyphal/transport/types.hpp#L58-L79

We discussed that inheritance is not a good choice here, so perhaps we should use composition. After some consideration I realized that the current arrangement will require much manual tinkering in the presentation layer and above.

```c++
struct TransferMetadata final
{
    TransferId transfer_id{};
    TimePoint  timestamp;
    Priority   priority{};
};

struct MessageTransferMetadata final
{
    TransferMetadata       base;
    cetl::optional<NodeId> publisher_node_id;
};

struct ServiceTransferMetadata final
{
    TransferMetadata base;
    NodeId           remote_node_id{};
};
```

serges147:
Closing as it has been fixed by pr #374

PR #374, opened 2024-08-08
CAN transport example

serges147:
- Added new `example_03_linux_socketcan_transport` target - appears only on Linux (but not on Mac f.e.)
- Eliminated `IRunnable`
- Addressed "Reduce duplication around TransferMetadata" issue #365
- Addressed "Refactor the execution logic" issue #367 
- Fixed all unit tests affected by executor vs runnable api changes
- Sonar fixes
- Latest CAVL


include/libcyphal/common/cavl/cavl.hpp
```diff
@@ -29,6 +29,7 @@
 #include <cstdint>
 #include <tuple>
 #include <type_traits>
+#include <utility>
```
serges147, line 32: This file can be skipped from review b/c it was reviewed at the original `pavel-kirienko/cavl` repo.

test/unittest/common/cavl/test_cavl.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <libcyphal/common/cavl/cavl.hpp>
 
+#include <gmock/gmock.h>
```
serges147, line 5: This file can be skipped from review b/c it was reviewed at the original pavel-kirienko/cavl repo.

include/libcyphal/transport/can/media.hpp
```diff
@@ -110,6 +110,30 @@ class IMedia
     CETL_NODISCARD virtual PopResult::Type pop(const cetl::span<cetl::byte> payload_buffer) noexcept = 0;
     ///@}
 
+    /// @brief Registers "ready to push" callback function at a given executor.
```
thirtytwobits, line 113: What does "ready to push" mean precisely. That is, ready to push what? An MTU-worth? one-byte?

thirtytwobits (commented):
Can we make clang-tidy optionatl please? It's a pain that I can't build when it's not available.

docs/examples/example_03_linux_socketcan_transport.cpp
```diff
@@ -0,0 +1,319 @@
...
+using testing::VariantWith;
+
+// NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
+
+class Example_03_LinuxSocketCanTransport : public testing::Test
+{
+protected:
+    void SetUp() override
+    {
+        std::cerr.imbue(std::locale("en_US.UTF-8"));
```
thirtytwobits, line 81: These throw on osx. I know this is a Linux example but we're doing the same thing over on some posix examples that otherwise run fine on mac. Can we just omit the imbue calls?

thirtytwobits (changes requested):

docs/examples/example_03_linux_socketcan_transport.cpp
serges147: I did it just for nice formatting of integers printed to console. ok, I'll remove them.

include/libcyphal/transport/can/media.hpp
serges147: It means that IO output is ready to accept next frame for transmission. In case of posix implementation it means that socketcan file descriptor (passed to POSIX `::poll`) becomes "ready" (namely `POLLOUT` flag is set at `revents`). See also `ITxSocket::registerCallback` docs.

But if you have f.e. CAN media at micro controller then it probably means that some interrupt had happened (signaling that CAN hw tx queue is done transmitting previous frame, and "ready" to accept more).

serges147:
> Can we make clang-tidy optionatl please? It's a pain that I can't build when it's not available.

There is cmake `-DNO_STATIC_ANALYSIS=YES` for that.

include/libcyphal/transport/can/media.hpp
thirtytwobits: Can we update the verbiage where we discuss "ready to push" in the code to indicate this is about ready to push frames instead of just saying "data"?

thirtytwobits (approved):

pavel-kirienko (approved):
I think you forgot to remove `udpGetSocketOptionInt`/`udpSetSocketOptionInt`

Issue #373, opened 2024-08-06
SingleThreadedExecutor: Consider removing pointer/int conversions


pavel-kirienko:
See https://github.com/OpenCyphal-Garage/libcyphal/pull/371#discussion_r1704756790_

One posisble way to approach this is to move the custom executor extension point into Callback::Any. This change affects only the SingleThreadedExecutor and everything below it, not the main codebase.

pavel-kirienko:
This is not a high-priority issue because it affects only the specific implementation of one Executor policy. The core library is not affected. It is already possible to create an execution policy that does not need this.

serges147:
Closing as fixed by PR #376

Issue #372, opened 2024-08-06
Remove `None` as a scheduling option


pavel-kirienko:
>This special case doesn't look convincing. I understand that there is some value in being able to unschedule a callback but as far as I understand by now we haven't identified an explicit need for that, and it would be easy to retrofit -- if needed -- later. The troubling thing here is that an extension of any well-structured set of entities with a special case often tends to break the design in unexpected ways at some point.
>
>>*But I couldn’t resist the temptation to put in a null reference, simply because it was so easy to implement.*

_Originally posted by @pavel-kirienko in https://github.com/OpenCyphal-Garage/libcyphal/pull/371#discussion_r1704753104_

serges147:
- The case of callback schedule cancellation could be covered by adding (in the future, if needed) by having new `callback.cancel()` method - callback will remain fully function but just in a dormant state, ready to be resheduled.
- `executor.scheduleCallback(callback, ...)` -> `callback.schedule(...)` move will eliminate another point of failure which is currently described in docs as "_...must be used only with the same executor..._"

serges147:
Closing as fixed by PR #376

PR #376, opened 2024-08-12
Remove None as a scheduling option

serges147:
Fixes for issues #372 & #373.

- Callback changes:
  - Eliminated `Callback::Schedule::None` alternative. Optional schedule now is stored instead.
  - Added new `Callback::Arg` struct - parameter of the `Callback::Function`. Contains:
    - `exec_time` - scheduled execution time of the callback
    - `approx_now` - actual (approximated) time of the execution
  - Introduced `Callback::Interface`
    - has `bool schedule(const Schedule::Variant&)` method.
- `IExecutor` changes:
  - Removed `scheduleCallback` method. Now scheduling done via callback interface.
- `SingleThreadedExecutor` (and its derived `PosixSingleThreadedExecutor`) don't rely anymore on unsafe `Callback::Handle` and `reinterpret_cast`-s. Strong typed `Callback::Interface` is in use instead.

Also:
- simplified RTTI usage (give up usage of `cetl::rtti_helper` which is not really helpful)
- other minor fixes (based on previous PR review notes)
- latest toolshed v 22.4.8

Coverage:
Screenshot from 2024-08-12 12-22-13

### DIFF
--- a/docs/examples/platform/posix/posix_single_threaded_executor.hpp
+++ b/docs/examples/platform/posix/posix_single_threaded_executor.hpp
@@ -136,9 +136,7 @@ public:
                 {
                     if (auto* const cb_interface = callback_interfaces_[index])
                     {
-                        const bool result = cb_interface->schedule(Callback::Schedule::Once{now_time});
-                        (void) result;
-                        CETL_DEBUG_ASSERT(result, "Failed to schedule callback.");
+                        cb_interface->schedule(Callback::Schedule::Once{now_time});
                     }
                 }
             }
--- a/docs/examples/platform/posix/posix_single_threaded_executor.hpp
+++ b/docs/examples/platform/posix/posix_single_threaded_executor.hpp
@@ -44,7 +44,7 @@ public:
         : total_awaitables_{0}
         , awaitable_nodes_allocator_{&memory_resource}
         , poll_fds_{&memory_resource}
-        , callback_handles_{&memory_resource}
+        , callback_interfaces_{&memory_resource}
     {
     }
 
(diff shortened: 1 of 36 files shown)


docs/examples/platform/posix/posix_single_threaded_executor.hpp
```diff
@@ -87,13 +87,13 @@ class PosixSingleThreadedExecutor final : public libcyphal::platform::SingleThre
         // so we can reuse them, and grow on demand (but never shrink).
         //
         poll_fds_.clear();
```
pavel-kirienko, line 89: I think it's worth adding a comment (considering that this is an example) that one could possibly avoid the need to traverse the awaitables on every poll by switching to another multiplexing API that stores fds in a kernel space structure, like epoll. Each awaitable item would then register and unregister itself with that kernel space object as needed and the poll would go straight to the blocking call without the need to reconstruct anything.

pavel-kirienko (commented):
We should be removing `MonotonicClock::now()` because we now have `IExecutor::now()`.

We should also talk about the onCallbackHandling/AwaitableNode issue offline.

serges147:
> We should be removing `MonotonicClock::now()` because we now have `IExecutor::now()`.

Yup, I remember about this - I wanted to introduce `ITimeProvider` in a separate PR. I believe it's better to have `now()` in separate interface (b/c one could be interested only in this functionality but not whole executor), and `IExecutor` will inherit from such `ITimeProvider`.

docs/examples/platform/posix/posix_single_threaded_executor.hpp
serges147: I would like (a bit later, separate dedicated PR) to do such exercise - switch to `::epoll`.

pavel-kirienko (approved):

lydia-at-amazon (approved):

PR #383, opened 2024-09-08
RPC Client implementation

serges147:
1. Implemented raw and strong-typed clients, and their `ResponsePromise`-s.
2. Moved some generic error under `libcyphal` namespace.

Example "example_1_presentation_1_ping_user_service" output:

image


include/libcyphal/transport/udp/svc_rx_sessions.hpp
```diff
@@ -132,7 +145,7 @@ class SvcRxSession final : IRxSessionDelegate, public Interface_  // NOSONAR cpp
     void setTransferIdTimeout(const Duration timeout) override
     {
         const auto timeout_us = std::chrono::duration_cast<std::chrono::microseconds>(timeout);
-        if (timeout_us.count() > 0)
+        if (timeout_us.count() >= 0)
```
pavel-kirienko, line 148: The idiomatic form is actually `duration::zero()`:

```suggestion
        if (timeout >= Duration::zero())
```

include/libcyphal/transport/transfer_id_generators.hpp
```diff
@@ -0,0 +1,141 @@
...
+    explicit SmallRangeTransferIdGenerator(const TransferId modulo) noexcept
+        : modulo_{modulo}
+        , next_transfer_id_{0}
+        , in_use_transfer_ids_{}
+    {
+        CETL_DEBUG_ASSERT(modulo > 0, "Transfer ID modulo must be greater than 0.");
+        CETL_DEBUG_ASSERT(modulo <= Size, "Transfer ID modulo must be <= than `Size`.");
+    }
+
+    /// @brief Returns the next available (not in use) transfer ID.
```
pavel-kirienko, line 93: Please document that the worst-case complexity is linear of the number of pending requests.

include/libcyphal/transport/transfer_id_generators.hpp
```diff
@@ -0,0 +1,141 @@
...
+/// Not supposed to be used directly by the users of the library.
+///
+namespace detail
+{
+
+/// @brief Defines a trivial transfer ID generator.
+///
+/// The generator is trivial in the sense that it simply modulo increments the transfer ID.
+/// B/c modulo is expected to be quite big (like >= 2^48), collisions of transfer ids are unlikely.
+/// Normally in use for UDP transport, where the modulo is `2^64 - 1`.
```
pavel-kirienko, line 33: Do we really need to bother with the modulo though? $2^{48}$ is already large enough to be unreachable for all practical scenarios, so we can just use plain increments.

include/libcyphal/presentation/errors.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <cetl/pf17/cetlpf.hpp>
+
+#include <nunavut/support/serialization.hpp>
+
+namespace libcyphal
+{
+namespace presentation
+{
+
+/// @brief Defines terminal 'expired' error state of the response promise.
```
pavel-kirienko, line 21: Let us please keep these highly specific error states next to their place of use. I think the promise-related errors should go right next to the Client definition, as they effectively form its API.

docs/examples/platform/posix/posix_platform_error.hpp
```diff
@@ -27,7 +27,7 @@ class PosixPlatformError final : public libcyphal::transport::IPlatformError
     {
         CETL_DEBUG_ASSERT(err > 0, "");
     }
-    virtual ~PosixPlatformError() noexcept                       = default;
+    virtual ~PosixPlatformError()                                = default;
```
pavel-kirienko, line 30: I see that you've changed quite a few dtors in this PR to be non-noexcept, but I don't see the point because throwing from a dtor is UB.

serges147: AFAIK it's redundant to specify `noexcept` at destructors - they all noexcept by default. Please correct me if I'm wrong. Thanks!

pavel-kirienko: ah yes you're right

include/libcyphal/presentation/errors.hpp
pavel-kirienko: No, I meant that they should be next to the Client class definition but not part of it

serges147: ok, I can do it, although I have 2 points more:
1. strictly speaking these two are more related to promise rather than client - hence the "ResponsePromised" prefix
2. there is kinda pattern at other layers of libcyphal, namely each layer has their own `<layer>/errors.hpp` which holds definitions of free standing errors and failures (variants).

docs/examples/platform/posix/posix_platform_error.hpp
serges147: btw, here is interesting commit from the Bjarne : https://github.com/isocpp/CppCoreGuidelines/commit/d3c3e3aa986fc9cf6238df2a30a3e071a2f55edd#comments

include/libcyphal/presentation/errors.hpp
pavel-kirienko: So are we moving them next to the promise? I get the idea of moving generic error types like argument error or memory error into an errors.hpp, but more specialized types should be kept close to the entity they pertain to.

serges147: done

pavel-kirienko (approved):

include/libcyphal/presentation/response_promise.hpp
```diff
@@ -0,0 +1,425 @@
...
+/// Raw response promise failure state could be only expired. In contrast see `ResponsePromiseFailure`,
+/// where the set of possible failure states is extended with additional points of failures.
+///
+using RawResponsePromiseFailure = cetl::variant<  //
+    ResponsePromiseExpired>;
+
+/// @brief Defines terminal failure state of the strong-typed response promise.
+///
+/// In addition to the raw failure states, this type also includes possible memory allocation errors,
+/// aa well as errors from the `nunavut` library in case of response deserialization issues.
```
lydia-at-amazon, line 54: nit: `aa` --> `as`

docs/examples/1_presentation/example_1_presentation_1_ping_user_service_udp.cpp
```diff
@@ -220,99 +263,96 @@ TEST_F(Example_1_Presentation_1_PingUserService_Udp, main)
...
+    state.transport_->setTransientErrorHandler(CommonHelpers::Udp::transientErrorReporter);
 
     // 2. Create presentation layer object.
     //
     Presentation presentation{mr_, executor_, *state.transport_};
 
     // 3. Bring up "Ping" server.
     //
-    auto maybe_ping_server = presentation.makeServer<UserService::PingRequest, UserService::PongResponse>(  //
+    // Fot the sake of demonstration, we will keep track of all "Ping" requests inside the `ping_contexts` map,
```
lydia-at-amazon, line 274: nit sp: `Fot` --> `For`

lydia-at-amazon (approved):

include/libcyphal/presentation/presentation.hpp
```diff
@@ -326,48 +335,124 @@ class Presentation final : private detail::IPresentationDelegate
...
 
+    /// @brief Makes a custom typed RPC client bound to a specific server node and service ids.
+    ///
+    /// @tparam Request The request type of the client. See `Client<Request, ...>` for more details.
+    /// @tparam Response The response type of the client. See `Client<..., Response>` for more details.
+    /// @param server_node_id The server node ID to bind this client with.
+    /// @param service_id The service ID of the client.
+    ///
+    template <typename Request, typename Response>
+    auto makeClient(const transport::NodeId server_node_id,
```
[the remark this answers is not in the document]
serges147: Yes, that is correct. I'll take a note, and add a bit more docs about that in the next pr.

PR #384, opened 2024-09-16
New "example_1_presentation_3_hb_getinfo_ping_linux_can" example app; some todos.

serges147:
- Addressed several TODOs:
  - // TODO: Switch to `transport::PayloadFragments`.
  - // TODO: Move `tryDeserialize` method to some common place
- Added new `TestCanSvcRxSessions::receive_multiple_tids_frames` - reproduces exactly what is described at the https://github.com/OpenCyphal/libcanard/issues/228
- Added new `example_1_presentation_3_hb_getinfo_ping_linux_can.cpp` example app - it's similar to `example_1_presentation_1_ping_user_service_udp.cpp` but with CAN transport.


include/libcyphal/presentation/common_helpers.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+/// Internal implementation details of the Presentation layer.
+/// Not supposed to be used directly by the users of the library.
+///
+namespace detail
+{
+
+using DeserializationFailure = cetl::variant<MemoryError, nunavut::support::Error>;
+
+template <typename Message>
```
pavel-kirienko, line 31: Interesting, we seem to be one template specialization away from closing this todo.

docs/examples/platform/common_helpers.hpp
```diff
@@ -50,6 +51,15 @@ struct CommonHelpers
         return iface_addresses;
     }
 
+    static std::string joinInterfaceAddresses(const std::vector<std::string>& iface_addresses)  // NOLINT
+    {
+        // join the interface addresses into a single string using algorithms.
```
pavel-kirienko, line 56: image

docs/examples/1_presentation/example_1_presentation_3_hb_getinfo_ping_linux_can.cpp
```diff
@@ -0,0 +1,443 @@
...
+    });
+    //
+    constexpr auto hb_period = std::chrono::seconds{uavcan::node::Heartbeat_1_0::MAX_PUBLICATION_PERIOD};
+    publish_heartbeat_every_1s_cb.schedule(Callback::Schedule::Repeat{startup_time_ + hb_period, hb_period});
+
+    // 7. Bring up 'GetInfo' server.
+    //
+    using GetInfo_1_0 = uavcan::node::GetInfo_1_0;
+    uavcan::node::GetInfo_1_0::Response get_info_response{{1, 0}};
+    const std::string                   node_name{"org.opencyphal.Ex_1_Pres_3_HB_GetInfo_Ping_CAN"};
```
pavel-kirienko, line 403: Can we minimize usage of heapy containers please so that the embedded types feel at home here? I see that string_view is not available in C++14 so maybe plain `const char* const` would work?

pavel-kirienko (approved):

docs/examples/1_presentation/example_1_presentation_3_hb_getinfo_ping_linux_can.cpp
serges147: It just examples. I use also vector, map, ostream - but none of these at production libcyphal code.

lydia-at-amazon: A bit cheap of me to pile on, but +1 to trying to minimize containers as most firmware don't have access to `std::vector` or `std::unordered_map`. Even though these are examples, examples are the biggest way that potential customers decide if a solution works from them so it might drive more embedded customers away. Probably fine to leave for now though, and thank you for making them!

lydia-at-amazon:
I see we fixed https://github.com/OpenCyphal/libcanard/issues/228, thank you!

lydia-at-amazon (approved):

docs/examples/1_presentation/example_1_presentation_3_hb_getinfo_ping_linux_can.cpp
serges147: Hmm... even `std::vector`?! I'm surprised b/c Nunavut's code generation is based on the vector (but of course based on PMR allocator).

serges147:
> I see we fixed [OpenCyphal/libcanard#228](https://github.com/OpenCyphal/libcanard/issues/228), thank you!

There is another small PR (draft yet) where I take advantage of the fix.

docs/examples/1_presentation/example_1_presentation_3_hb_getinfo_ping_linux_can.cpp
lydia-at-amazon: I think our firmware applications typically use the C headers, which don't use `std::vector`

PR #388, opened 2024-10-07
First draft of Registry

serges147:
Please note that target branch is not `main` but `issue/registry` - this first PR is just initial commit to the whole "registry" topic/feature - not yet workable. My idea is keep feature under `issue/registry` until it will be ready functional wise, and then post it for Amazon review when it start working (including examples).


include/libcyphal/application/node/heartbeat_producer.hpp
```diff
@@ -121,10 +121,10 @@ class HeartbeatProducer final  // NOSONAR cpp:S4963
...
     ///
     /// @param update_callback_fn The update callback function, which will be called before publication of the next
-    ///                           heartbeat message. Allows to modify the `message` before it will be published.
-    ///                           `arg.message.uptime` field is automatically prepopulated to reflect duration
-    ///                           since the node startup, so although the field could be modified by the user,
-    ///                           it will be anyway overridden on the next update.
+    ///                           heartbeat message. Allows the user to modify the `message` before it will be
+    ///                           published. `arg.message.uptime` field is automatically prepopulated to reflect
+    ///                           duration since the node startup, so although the field could be modified by the user,
+    ///                           it will be overridden anyway on the next update.
```
serges147, line 127: previous pr review grammar fixes

docs/examples/1_presentation/example_1_presentation_1_ping_user_service_udp.cpp
```diff
@@ -25,6 +25,8 @@
 #include <libcyphal/transport/udp/udp_transport.hpp>
 #include <libcyphal/transport/udp/udp_transport_impl.hpp>
 #include <libcyphal/types.hpp>
+
+#include <cassert>  // NOLINT for NUNAVUT_ASSERT
```
pavel-kirienko, line 29: I recall it was discussed somewhere that `serialization.hpp` needs to `#include <cassert>` so that the user code doesn't have to do it. We don't have an issue about this do we? If not, please make one.

include/libcyphal/common/crc.hpp
```diff
@@ -0,0 +1,138 @@
...
+        // No lint for cppcoreguidelines-avoid-magic-numbers and readability-magic-numbers.
+        // Also ignore cppcoreguidelines-pro-bounds-constant-array-index - we are using a lookup table.
+        // NOLINTNEXTLINE
+        crc_ = (CrcTable()[b ^ (crc_ >> 56U)] ^ (crc_ << 8U));
+    }
+
+    std::uint64_t crc_ = std::numeric_limits<std::uint64_t>::max();
+
+    // No lint for cppcoreguidelines-avoid-magic-numbers and readability-magic-numbers.
+    static const std::array<std::uint64_t, 256>& CrcTable() noexcept  // NOLINT
```
pavel-kirienko, line 62: camelCase please; also the field above should go after the methods

include/libcyphal/application/registry/register_impl.hpp
```diff
@@ -0,0 +1,350 @@
...
+    const Name            name_;
+    const Options         options_;
+
+};  // RegisterBase
+
+// MARK: -
+
+/// Defines a register implementation template.
+///
+template <typename Getter, typename Setter, bool IsMutable>
```
pavel-kirienko, line 109: Unless I'm misreading it, IsMutable seems to be redundant because you already have a specialization on `Setter=void` for that purpose

include/libcyphal/application/registry/register_impl.hpp
```diff
@@ -0,0 +1,350 @@
...
+};  // Register<IsMutable=true>
+
+// MARK: -
+
+/// Defines a parameter-based register implementation template.
+///
+/// Instead of "external" getter/setter functions, the register uses member value storage.
+///
+template <typename ValueType, bool IsMutable = true>
+class ParamRegister final : public RegisterBase
```
pavel-kirienko, line 229: We seem to be missing callable registers that don't store the value but instead construct it ad-hoc; in the original codebase it was done via `route`: https://github.com/Zubax/telega/blob/80f5203d8536b7be7e6fc991a8462b73ea5930cb/telega/src/registry/registry_impl.hpp#L151-L197

This is an essential feature

include/libcyphal/application/registry/registry_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+    IRegister::Name index(const std::size_t index) const override
+    {
+        if (const auto* const reg = registers_tree_[index])
+        {
+            return reg->getName();
+        }
+        return IRegister::Name{};
+    }
+
+    bool append(IRegister& reg) override
```
pavel-kirienko, line 83: Why can't we make factory methods here that create a register instance and then add it in one operation?

docs/examples/1_presentation/example_1_presentation_1_ping_user_service_udp.cpp
serges147: Created https://github.com/OpenCyphal/nunavut/issues/348

include/libcyphal/application/registry/registry_impl.hpp
serges147: Done - see `Registry::route`, `Registry::expose` and `Registry::exposeParam` methods.
Please note that I intentionally left my original `makeRegister` methods as is b/c:
- `route` overload methods use them
- user could use them if he wants:
  - first **infallibly** create registers (f.e. as members during constructor),
  - and only later link them to the registry (which might fail due to name duplicates).

Added more docs mentioning all possible factory alternatives.

include/libcyphal/application/registry/register_impl.hpp
serges147: See below comment

include/libcyphal/application/registry/register.hpp
```diff
@@ -0,0 +1,188 @@
...
+        /// True if the register value is retained across application restarts.
+        bool persistent_{false};
+    };
+
+    /// Defines a pair of the register value and its flags.
+    ///
+    struct ValueAndFlags final
+    {
+        Value value_;
+        Flags flags_;
```
pavel-kirienko, line 59: Public fields should not have the underscore suffix

include/libcyphal/application/registry/registry_impl.hpp
```diff
@@ -0,0 +1,215 @@
...
+        return !std::get<1>(register_existing);
+    }
+
+    // MARK: - Other factory methods:
+
+    /// Constructs a new read-only register, and links it to this registry.
+    ///
+    /// @param name The name of the register. Should be unique within the registry.
+    /// @param options Extra options for the register, like "persistent" option (`true` by default).
+    /// @param getter The getter function to provide the register value (from a storage).
```
pavel-kirienko, line 109:
```suggestion
    /// @param getter The getter function to provide the register value.
```

There may be no storage at all. It is common to construct values ad hoc.

include/libcyphal/application/registry/registry_impl.hpp
```diff
@@ -0,0 +1,215 @@
...
+        CETL_DEBUG_ASSERT(std::get<0>(register_existing)->isLinked(), "Should be linked.");
+        return !std::get<1>(register_existing);
+    }
+
+    // MARK: - Other factory methods:
+
+    /// Constructs a new read-only register, and links it to this registry.
+    ///
+    /// @param name The name of the register. Should be unique within the registry.
+    /// @param options Extra options for the register, like "persistent" option (`true` by default).
```
pavel-kirienko, line 108:
```suggestion
    /// @param options Extra options for the register, like "persistent" option (`false` by default).
```

include/libcyphal/application/registry/registry_impl.hpp
```diff
@@ -0,0 +1,215 @@
...
+        }
+        return cetl::nullopt;
+    }
+
+    /// Constructs a new read-write register, and links it to this registry.
+    ///
+    /// @param name The name of the register. Should be unique within the registry.
+    /// @param options Extra options for the register, like "persistent" option (`true` by default).
+    /// @param getter The getter function to provide the register value (from a storage).
+    /// @param setter The setter function to update the register value (at the storage).
```
pavel-kirienko, line 130:
```suggestion
    /// @param options Extra options for the register, like "persistent" option (`false` by default).
    /// @param getter The getter function to provide the register value.
    /// @param setter The setter function to update the register value.
```

include/libcyphal/application/registry/registry_impl.hpp
```diff
@@ -0,0 +1,215 @@
...
+    /// @return Result register if it was appended successfully. Otherwise, `nullopt`.
+    ///
+    template <typename Getter, typename Setter>
+    cetl::optional<Register<Getter, Setter>> route(const IRegister::Name     name,
```
pavel-kirienko, line 134: Can we please also keep the simple route overload that takes only the name and the getter? Read-only non-persistent registers are very common so the helper is, well, helpful.

    /// Convenience method for creating getters with default options.
    template <NameConcept N, Getter G>
    [[nodiscard]] RegisterPtr route(N&& name, G&& getter)
    {
        return add(std::forward<N>(name), Options{}, std::forward<G>(getter), std::monostate{});
    }

include/libcyphal/application/registry/registry_impl.hpp
```diff
@@ -0,0 +1,215 @@
...
+    /// In contrast to the above `expose()`, this method allows one to expose a parameter with a default value.
+    /// Exposed parameter value is stored inside the register and can be mutated (by default).
+    ///
+    /// @param name The name of the register. Should be unique within the registry.
+    /// @param default_value Initial default value.
+    /// @param options Extra options for the register, like "persistent" option.
+    /// @return Result register if it was appended successfully. Otherwise, `nullopt`.
+    ///
+    template <typename T, bool IsMutable = true>
+    cetl::optional<ParamRegister<T, IsMutable>> exposeParam(const IRegister::Name     name,
```
pavel-kirienko, line 181: Not sure I understand the purpose of this. Isn't this equivalent to:

```
T value = default;
auto register = registry.expose("name", value);
```

If this is meant to model the `P` class in Telega, then I think it shouldn't be here in this abstract registry module because it is too application-specific.

pavel-kirienko (changes requested):

include/libcyphal/application/registry/registry_impl.hpp
serges147: Yes, it meant to model the `P`. I don't see it as application-specific. In contrast, I think it allows to expose a value as a register without worrying where the value is stored and its lifetime relative to the register - very light IMO, and without needed to store also 2 lambdas.

include/libcyphal/application/registry/registry_impl.hpp
serges147: What do you think about changing order of `options` and `getter`/`setter` lambdas. Like:
```c++
    template <typename Getter>
    cetl::optional<Register<Getter, void>> route(const Name name,
                                                 Getter&& getter,
                                                 const IRegister::Options& options = {});
    template <typename Getter, typename Setter>
    cetl::optional<Register<Getter, void>> route(const Name name,
                                                 Getter&& getter,
                                                 Setter&& setter,
                                                 const IRegister::Options& options = {});
```
In this way we can have `options` parameter with `{}` default? In Swift language it's actually very recommended to have closures (aka lambdas) as last parameters, but it is mostly b/c of very nice "terminal closure" syntax Swift has... But for C++ I'm not aware of any recommendation about the order.

include/libcyphal/application/registry/register.hpp
serges147: will fix in the next pr

include/libcyphal/application/registry/registry_impl.hpp
pavel-kirienko: Sounds good.

include/libcyphal/application/registry/registry_impl.hpp
pavel-kirienko: Ok, then we need a different name for the method: not `exposeParam` but something like just `param[eter]`.

pavel-kirienko (approved):
Tentative approval with some pending changes.

include/libcyphal/application/registry/registry_impl.hpp
serges147: >param
hmm... what about a verb in the method's name?
I skip verb mostly in very simple cases only - like returning just reference to a private member (see f.e. `Node::heartbeatProducer() -> node::HeartbeatProducer&`). What about `makeParam`?

pavel-kirienko: `makeParam` is okay but then maybe `parameterize`?

serges147: ok, I'll go with "parameterize"
in the next PR

PR #392, opened 2024-10-30
Simplified registry implementation

serges147:
Simplified registry implementation by...
- removing user specific coercion logic - now API based on either the Nunavut's `Value` type or one of its variants;
- only `Registry::route` method left - `expose` and `parameterize` were removed (b/c they depend on the coercion)


include/libcyphal/application/registry/register.hpp
```diff
@@ -23,19 +30,56 @@ namespace registry
...
+    ///
+    Mutability,
+
+    /// Value cannot be coerced to the register type.
+    ///
+    /// Libcyphal in not using this error directly, but user's setter might return it
+    /// to indicate that a new value cannot be accepted due to an incompatible type of the value.
+    ///
+    Coercion,
+
```
pavel-kirienko, line 49: This is no longer used

include/libcyphal/types.hpp
```diff
@@ -57,6 +57,12 @@ template <typename Interface, typename Any>
 class ImplementationCell final
 {
 public:
+    ImplementationCell()
+        : fn_getter_mut_{[](Any&) -> Interface* { return nullptr; }}
+        , fn_getter_const_{[](const Any&) -> const Interface* { return nullptr; }}
+    {
+    }
+
```
pavel-kirienko, line 65: Is there no way to avoid this? The absence of the default ctor is advantageous because it reduces the risk of having an instance of the implementation cell in an invalid state. The only way to get it to be invalid is to move the implementation out of it, which is only possible if Any is movable.

pavel-kirienko (approved):

include/libcyphal/types.hpp
serges147: I'll take a look.

include/libcyphal/application/registry/register.hpp
serges147: ok, I will use `Semantics` instead

include/libcyphal/types.hpp
serges147: deleted - thank you for noticing!

include/libcyphal/application/registry/register.hpp
```diff
@@ -173,6 +218,43 @@ class IRegister : public cavl::Node<IRegister>  // NOSONAR cpp:S4963
 
 };  // IRegister
 
+/// Defines type-erased register.
+///
+/// The Footprint size of the unbounded variant should be enough for any register implementation.
+/// The implementation should not be copyable but moveable.
+///
+template <std::size_t Footprint>
+using Register = ImplementationCell<IRegister, cetl::unbounded_variant<Footprint, false, true>>;
```
lydia-at-amazon, line 227: This is a bit late in the game, but do you think that we've made the Libcyphal code more complicated by implementing type erasure using the ImplementationCell? In other words, should we have just used Templates since we're using them elsewhere? I think maybe this is something we could talk about in the dev call

lydia-at-amazon (approved):

include/libcyphal/application/registry/register.hpp
lydia-at-amazon: Nevermind, Scott allayed my concerns :)

PR #393, opened 2024-11-01
Defined `platform::storage::IKeyValue` interface

serges147:
- Defined abstract key-value storage interface; plus its mock under unit tests.
- Added `load` & `save` methods for persisting registry in a storage; its unit test coverage (100%).
- Extended example_2_app_0 with registry persisting.

### DIFF
--- a/docs/examples/platform/storage.hpp
+++ b/docs/examples/platform/storage.hpp
@@ -1,0 +1,120 @@
+/// @copyright
+/// Copyright (C) OpenCyphal Development Team  <opencyphal.org>
+/// Copyright Amazon.com Inc. or its affiliates.
+/// SPDX-License-Identifier: MIT
+
+#ifndef EXAMPLE_PLATFORM_STORAGE_HPP_INCLUDED
+#define EXAMPLE_PLATFORM_STORAGE_HPP_INCLUDED
+
+#include <cetl/pf17/cetlpf.hpp>
+#include <cetl/pf20/cetlpf.hpp>
+#include <libcyphal/platform/storage.hpp>
+#include <libcyphal/types.hpp>
+
+#include <cerrno>
+#include <cstddef>
+#include <fstream>
+#include <stdio.h>
+#include <string>
+#include <string_view>
+#include <sys/stat.h>
+#include <utility>
+
+namespace example
+{
+namespace platform
+{
+namespace storage
+{
+
+class KeyValue final : public libcyphal::platform::storage::IKeyValue
+{
+public:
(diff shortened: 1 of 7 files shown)


docs/examples/platform/storage.hpp
```diff
@@ -0,0 +1,120 @@
...
+        {
+            std::cerr << "Error making folder: '" << root_path_ << "'.\n";
+            std::cerr << "Error: " << std::strerror(errno) << std::endl;
+        }
+    }
+
+private:
+    using Error = libcyphal::platform::storage::Error;
+
+    std::string makeFilePath(const cetl::string_view key) const
```
pavel-kirienko, line 46: In practice I would expect the keys to be hashed, so it won't be necessary to deal with directory nesting. This is fine for a demo but it might be worth noting in a comment that since we don't need key listing, we don't have to retain the key names, which greatly simplifies things.

include/libcyphal/application/registry/registry_impl.hpp
```diff
@@ -159,6 +164,201 @@ class Registry final : public IIntrospectableRegistry
...
+        if (const auto err = action(reg_name))
+        {
+            return err;
+        }
+    }
+    return cetl::nullopt;
+}
+
+inline auto handleKeyValueGet(const platform::storage::IKeyValue& kv,
+                              IIntrospectableRegistry&            rgy,
```
lydia-at-amazon, line 206: nit: In PrimeAir's coding standard, we have a rule to always spell out all identifiers, so `kv` would be `key_value` and `rgy` would be `registry`. Fine to leave as is since we're following the Zubax coding standard, but just thought I'd call this out since I personally prefer the more verbose name :)

include/libcyphal/platform/storage.hpp
```diff
@@ -0,0 +1,94 @@
...
+    /// @param data The buffer to write the data to.
+    /// @return Either the number of bytes read or an error.
+    ///
+    virtual auto get(const cetl::string_view key, const cetl::span<std::uint8_t> data) const  //
+        -> Expected<std::size_t, Error> = 0;
+
+    /// Store data under a key.
+    ///
+    /// Existing data, if any, is replaced entirely.
+    /// New file and its parent directories created implicitly.
```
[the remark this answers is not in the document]
lydia-at-amazon: For example, to Pavel's comment, we store register values in the EEPROM for some of our firmware devices.

lydia-at-amazon (approved):

PR #394, opened 2024-11-05
Set default PMR

serges147:
Set default PMR, so that default behavior of `std::pmr::polymorphic_allocator` stays with specific memory resource (instead of default "new/delete" one) #verification #docs #sonar


thirtytwobits (approved):
Nice.

serges147:
ok, merging as there is at least one approve

PR #397, opened 2024-11-13
Implemented TODO to make asynchronous destruction of presentation entities

serges147:
Implemented TODO to make asynchronous destruction of presentation entities (publishers, subscribers and rpc clients).
Now unreferenced shared objects (with ref_count==0) will be quickly appended to the double-linked list, which will be traversed asynchronously in a near future callback (schedule for `now()`).


include/libcyphal/presentation/presentation.hpp
```diff
@@ -628,30 +673,57 @@ class Presentation final : private detail::IPresentationDelegate
...
 
     template <typename SharedNode>
-    void releaseSharedNode(SharedNode* const shared_node, cavl::Tree<SharedNode>& tree) const noexcept
+    static void forgetSharedNode(SharedNode& shared_node) noexcept
     {
-        CETL_DEBUG_ASSERT(shared_node != nullptr, "");
+        CETL_DEBUG_ASSERT(shared_node.isLinked(), "");
+        CETL_DEBUG_ASSERT(!shared_node.isReferenced(), "");
 
-        // TODO: make it async (deferred to "on idle" callback).
```
serges147, line 635: This is the "TODO" this PR is about.

lydia-at-amazon (approved):

include/libcyphal/presentation/shared_object.hpp
```diff
@@ -25,9 +25,53 @@ namespace presentation
...
+
+            next_node = nullptr;
+            prev_node = nullptr;
+        }
+    }
+
+    // No Lint b/c this `UnRefNode` is a simple helper struct as base of the below `SharedObject`.
+    // It's under `detail` namespace and not supposed to be used directly by the users of the library.
+    // NOLINTBEGIN(misc-non-private-member-variables-in-classes)
+    UnRefNode* prev_node{nullptr};
```
thirtytwobits, line 66: could we protect these by making them private then using `friend class SharedObject`?

include/libcyphal/presentation/presentation.hpp
```diff
@@ -112,7 +115,30 @@ class Presentation final : private detail::IPresentationDelegate
...
+        });
+        CETL_DEBUG_ASSERT(unref_nodes_deleter_callback_, "Should not fail b/c we pass proper lambda.");
+    }
+
+    Presentation(const Presentation&)                = delete;
+    Presentation(Presentation&&) noexcept            = delete;
+    Presentation& operator=(const Presentation&)     = delete;
+    Presentation& operator=(Presentation&&) noexcept = delete;
+
+    ~Presentation()
```
thirtytwobits, line 132: should this be virtual? I'm not sure TBH.

thirtytwobits (commented):
No blockers

include/libcyphal/presentation/presentation.hpp
serges147: `Presentation` is `final`, so I don't think we need `virtual`.

include/libcyphal/presentation/shared_object.hpp
serges147: we probably can but I'm not sure we need - all derived classes (from `SharedObject`) are still "Impl" internal (and under `detail` namespace) entities. F.e. if you work with `Publisher<Message>` (which is not shared object!) then you don't see under the hood involved `PublisherImpl`). The same story for 2 other entities: subscriber and rpc client.

PR #401, opened 2024-11-21
Sshirokov/352 zero copy

serges147:
- Added new `IMedia::getTxMemoryResource` method.
- Media implementations (at examples) now have their own TX memory resource.
- Addopted recent v4 canard api changes.
  - each `canardTxInit` now initialized with media TX memory resource
  - addressed "// TODO: Remove this workaround when the issue is resolved." - canard memory management now has de-allocation amount parameter.
- Extended all transport related unit tests to verify that TX pipeline uses its dedicated memory resource.


pavel-kirienko (approved):
Technically it's not zero-copy but single-copy, by design.

Issue #400, opened 2024-11-15
Publisher<Message>::publish puts message serialization buffer on the stack instead of using the allocators


emrainey:
For very large messages (i.e. 9KB+) and system with small stacks (i.e. 8KB), this practice can cause Stack problems. The message serialization buffer is used almost immediately in the liz-ards to make frame sequences so it would be a temporary large allocation. 

https://github.com/OpenCyphal-Garage/libcyphal/blob/main/include/libcyphal/presentation/publisher.hpp#L182

serges147:
Yup, you are right, for serialization we currently 3 such places. In contrast, for de-serialization we have `SmallPayloadSize` approach (see one single place at `tryDeserializePayload`).

@emrainey @pavel-kirienko What do you think guys, should we use similar `SmallPayloadSize` threshold in case of serialization? Although we can't say ahead for sure what would be result size of serialized message, but at least we know its maximum size (the `Message::_traits_::SerializationBufferSizeBytes`), so we can do something like this:
```
if (Message::_traits_::SerializationBufferSizeBytes <= SmallPayloadSize)) {
    // use current stack-based serialization
} else {
    // allocate temp buffer from PMR
}
```
OR, we can have two different template specializations of `Publisher`, `Client` and `Server` based on compile time `<= SmallPayloadSize` condition).

Other alternative could an extra `void setSmallPayloadThreshold(std::size_t threshold)` method (or two separate methods for TX and RX in case of RPC client/server), so that it will be decided at runtime (whether to use stack or dynamic memory), and potentially per entity (publisher, subscriber, client or server). Default value of these thresholds would be current `SmallPayloadSize`.

Please let me know your thoughts and preferences.

pavel-kirienko:
I think the comparison needs to be done at compile time, with specializations, because your runtime `if` may still cause stack allocation in the first branch even if it's not taken.

The threshold should be set to a reasonable small value that can be overridden via an optional preprocessor configuration parameter like `LIBCYPHAL_CONFIG_STACK_ALLOC_MAX`.

emrainey:
Scott indicated that there was a compromise where you could create a small allocators on the stack then pass it's upstream as the large allocator which is off the stack. If it fit in the small allocator, it would stay on the stack.

serges147:
>... you could create a small allocators on the stack

It's exactly what Pavel wanted to avoid, namely:
>... because your runtime if may still cause stack allocation in the first branch even if it's not taken.

[...]

Issue #398, opened 2024-11-14
Executor doesn't implement cetl::rtti


thirtytwobits:
For example: https://github.com/OpenCyphal-Garage/libcyphal/blob/main/include/libcyphal/executor.hpp#L108
We have the same methods defined but they are not virtual and the `cetl::rtti` interface is not used.

serges147:
The line you mention is not really about `IExecutor` but its `Callback::Interface`. But you are right - I need review all places where rtti is implemented. The reason why it's working currently (and causing any problems I'm aware of) is b/c `cetl::rtti_cast<>` implementation is not requiring currently to inherit from `cetl::rtti` interface - @pavel-kirienko Pavel, is it a **feature** or a **bug**? ;)
IMO on one hand it could be a **feature**, namely:
  - if you have very strict and predictable scenario from which type to what type you gonna `rtti_cast`
  - then you don't really need it be polymorphic - executor's callback stuff proves it (for better or worse)
On the other hand, it is probably not good that we have `rtti` interface defined but don't enforce it properly... f.e. if I add to the interface a new pure virtual method then I would expect to see compilation failures about classes being abstract...

pavel-kirienko:
That interface is mostly needed for the rtti_helper, so consider it a feature. We should implement that interface at least for regularity, though (making it `final` to avoid vtable lookup overheads where possible).

PR #408, opened 2024-12-05
`libcyphal::config`

serges147:
- Now sending uses stack for serialization of small messages/request/responses, and PMR otherwise (#400)
- Introduced `libcyphal::config`
- Fix for #398 

Also:
- eliminated deprecated NOSONAR cpp:S4963 - I disabled the rule at Sonar

### DIFF
--- a/include/libcyphal/common/cavl/cavl.hpp
+++ b/include/libcyphal/common/cavl/cavl.hpp
@@ -58,10 +58,9 @@ class Tree;
 /// The size of this type is 4x pointer size (16 bytes on a 32-bit platform).
 ///
 /// No Sonar cpp:S1448 b/c this is the main node entity without public members - maintainability is not a concern here.
-/// No Sonar cpp:S4963 b/c `Node` supports move operation.
 ///
 template <typename Derived>
-class Node  // NOSONAR cpp:S1448 cpp:S4963
+class Node  // NOSONAR cpp:S1448
 {
     // Polyfill for C++17's std::invoke_result_t.
     template <typename F, typename... Args>
@@ -954,8 +953,7 @@ private:
     /// the same call stack) we may occasionally fail to detect a bona fide case of a race condition, but this is
     /// acceptable because the purpose of this feature is to provide a mere best-effort data race detection.
     ///
-    /// No Sonar cpp:S4963 b/c of the RAII pattern.
-    class TraversalIndicatorUpdater final  // NOSONAR cpp:S4963
+    class TraversalIndicatorUpdater final
     {
     public:
         explicit TraversalIndicatorUpdater(const Tree& sup) noexcept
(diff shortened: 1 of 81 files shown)


include/libcyphal/common/cavl/cavl.hpp
pavel-kirienko, line 1: Should we upstream the changes?

pavel-kirienko, line 1: Should we upstream the changes?

include/libcyphal/config.hpp
```diff
@@ -0,0 +1,190 @@
...
+/// All methods are `static constexpr` - they are evaluated at compile time.
+///
+/// Nolint b/c this is the main purpose of this file - define various constants.
+/// NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
+///
+struct Config
+{
+    /// Defines max footprint of a callback function in use by the executor.
+    ///
+    static constexpr std::size_t IExecutor_Callback_FunctionMaxSize()  // NOSONAR cpp:S799
```
pavel-kirienko, line 29: Why are these functions instead of simple `static constexpr` constants? iirc even in C++14 these can be defined without out-of-line definitions.

test/unittest/custom_libcyphal_config.hpp
```diff
@@ -0,0 +1,58 @@
...
+{
+
+// Forward declaration of the custom configuration struct.
+struct MyConfig;
+
+}  // namespace custom
+
+// Overriding the default libcyphal configuration file with custom one.
+// NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
+#define LIBCYPHAL_CONFIG ::custom::MyConfig
```
pavel-kirienko, line 19: Shouldn't we use a more traditional approach like in o1heap?

```c++
#ifdef LIBCYPHAL_CONFIG_HEADER
#   include LIBCYPHAL_CONFIG_HEADER
#else
#   include "libcyphal/fallback_config.hpp"
#endif
```

include/libcyphal/config.hpp
serges147: ODR problem in c++14 if you pass such constant to somewhere which needs address of it (like `std::max` f.e.).

test/unittest/custom_libcyphal_config.hpp
serges147: ok, I'll switch gears a bit

serges147: Actually, I forgot that one of my design goals was to give client possibility to override only needed constants while automatically inheriting other defaults as is. Please see `test/unittest/custom_libcyphal_config.hpp` where I do `struct MyConfig : libcyphal::Config { ... }`.

pavel-kirienko (approved):

Issue #412, opened 2025-01-22
Only one single RPC client is possible per port.


serges147:
Multiple RPC clients, all on the same port but for different remote (server) node should be a valid scenario.
Currently, transport layer accepts the very first attempt to create Response RX session for a port,
but rejects all other consequent attempts with "already exists" failure for the same port.

- For **message** RX sessions, the described above "single rx session per port" behavior is desired one (presentation layer deals with it by allowing to have multiple `presentation::Subscriber` instances if needed).
- Even for service **request** RX sessions it's also desired (b/c only one server could be per port).
- But for service **response** RX sessions it's causing the problem (namely see title of this issue).

Problem found as part of OCVSMD work, where there was an attempt to create multiple **concurrent** RPC `uavcan/node/435.ExecuteCommand.1.3` clients, one per distinct node id, so that distribute a single command (like "COMMAND_RESTART" or "COMMAND_BEGIN_SOFTWARE_UPDATE") to multiple destination nodes.
Currently, the only workaround is execute such command in serial manner (create rpc client, send request, wait for response, release the client, repeat with next node).

- [x] Unit test for UDP transport
- [x] Fix UDP transport
- [x] Unit test for CAN transport
- [x] Fix CAN transport
- [x] Try at OCVSMD

serges147:
Fixed by PR #415

PR #415, opened 2025-01-24
Fix for issue #412 Only one single RPC client is possible per port

serges147:
1. Response Rx Session now doesn't own lizard's rpc port subscription - instead now session references the shared subscription (with reference counting).
2. Transport delegates now manage subscriptions.
3. CAN Response Rx Sessions now stored in cavl tree (for UDP it was already like this) - matching now by both port and node ids (previously it was by the port only).
4. CAN filters now also are made by the delegate (b/c it knows about subscriptions (see bullet # 2).
5. Extended unit tests to cover multiple response Rx sessions (both CAN and UDP). Also covered unsolicited responses.

Github's Hide whitespaces is recommended.

### DIFF
--- a/include/libcyphal/transport/can/delegate.hpp
+++ b/include/libcyphal/transport/can/delegate.hpp
@@ -6,6 +6,7 @@
 #ifndef LIBCYPHAL_TRANSPORT_CAN_DELEGATE_HPP_INCLUDED
 #define LIBCYPHAL_TRANSPORT_CAN_DELEGATE_HPP_INCLUDED
 
+#include "media.hpp"
 #include "rx_session_tree_node.hpp"
 
 #include "libcyphal/transport/errors.hpp"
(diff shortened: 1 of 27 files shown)


include/libcyphal/transport/udp/rx_session_tree_node.hpp
```diff
@@ -0,0 +1,121 @@
...
+{
+    UniquePtr<SocketInterface> interface;
+    IExecutor::Callback::Any   callback;
+
+};  // SocketState
+
+class IRxSessionDelegate;
+class IMsgRxSessionDelegate;
+
+struct RxSessionTreeNode
```
thirtytwobits, line 49: Missing class docs

include/libcyphal/transport/session_tree.hpp
```diff
@@ -0,0 +1,204 @@
...
+#include <array>
+#include <functional>
+#include <tuple>
+
+namespace libcyphal
+{
+namespace transport
+{
+
+/// Internal implementation details of the UDP transport.
```
thirtytwobits, line 26: The comment is wrong, right? This is no longer UDP-specific?

include/libcyphal/transport/session_tree.hpp
```diff
@@ -0,0 +1,204 @@
...
+        nodes_.traversePostOrder([this](auto& node) { destroyNode(node); });
+    }
+
+    CETL_NODISCARD bool isEmpty() const noexcept
+    {
+        return nodes_.empty();
+    }
+
+    template <bool ShouldBeNew = false, typename Params, typename... Args>
+    CETL_NODISCARD auto ensureNodeFor(const Params& params,
```
thirtytwobits, line 78: Please document this method.

include/libcyphal/transport/udp/delegate.hpp
```diff
@@ -220,7 +298,70 @@ class TransportDelegate
 
     CETL_NODISCARD UdpardRxRPCDispatcher& getUdpardRpcDispatcher() noexcept
     {
-        return rpc_dispatcher_;
+        return rx_rpc_dispatcher_;
+    }
+
+    void listenForRxRpcPort(UdpardRxRPCPort& rpc_port, const RequestRxParams& params)
```
thirtytwobits, line 304: you could avoid the code duplication here by specializing on a private template method with a "is_request" parameter.

thirtytwobits (commented):
Looking good. Mostly cosmetic comments.

pavel-kirienko (approved):

include/libcyphal/transport/can/delegate.hpp
```diff
@@ -196,16 +227,28 @@ class TransportDelegate
 
     struct SessionEvent
```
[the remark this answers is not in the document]
serges147: Normally, non-public stuff (under `detail` namespaces) is much less documented than public one. Anyway, I will add couple words. Thanks.

include/libcyphal/transport/udp/rx_session_tree_node.hpp
serges147: Same story here - everything about/under "delegate" is implementation details.

include/libcyphal/transport/can/delegate.hpp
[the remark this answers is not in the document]
thirtytwobits: Yeah, thanks. Even internal stuff should have some communication to future contributors about what the intended use is. The comments you provided are exactly what I was looking for. Thanks.

include/libcyphal/types.hpp
```diff
@@ -149,6 +150,31 @@ CETL_NODISCARD UpVariant upcastVariant(Variant&& variant)
...
+#if defined(__cpp_exceptions)
+    try
+    {
+#endif
+        std::forward<Action>(action)();
+        return true;
+#if defined(__cpp_exceptions)
+    } catch (const Exception& ex)
+    {
+        CETL_DEBUG_ASSERT(false, ex.what());
```
thirtytwobits, line 172: Sorry but I was thinking this was distinct from the debug assert. Failures may be recoverable and this would lower the utility of debug assert which should be used for "this should never happen" cases only. Uh...just remove this line and ignore my previous request. Sorry.

thirtytwobits (changes requested):

include/libcyphal/types.hpp
serges147: ok

thirtytwobits (approved):

PR #441, opened 2025-04-22
Fixes for issues 431 & 399

serges147:
- `libcyphal::MonotonicClock` type definition is moved under `libcyphal::Config` - now it's possible to customize libcyphal `TimePoint` & `Duration` types.
- Fixed several missing `std::chrono::duration_cast<Duration>` when converting from lizard µsec to libcyphal time.
- Added two new unit tests to ensure that both CAN and UDP transports could be compiled with custom milliseconds-based clock. Without the above `duration_cast` fixes these unit tests are not-compilable.

### DIFF
--- a/include/libcyphal/transport/can/can_transport_impl.hpp
+++ b/include/libcyphal/transport/can/can_transport_impl.hpp
@@ -578,8 +578,8 @@ private:
         }
         const IMedia::PopResult::Metadata& pop_meta = *pop_success;
 
-        const auto timestamp_us =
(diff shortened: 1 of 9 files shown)


include/libcyphal/transport/can/can_transport_impl.hpp
```diff
@@ -578,8 +578,8 @@ class TransportImpl final : private TransportDelegate, public ICanTransport
         }
         const IMedia::PopResult::Metadata& pop_meta = *pop_success;
```
thirtytwobits, line 579: Ah yes. To be successful in pop music one must do this.

thirtytwobits (approved):

